PR #6, opened 2019-01-04
add JSONExporter for Extract and Export

### LINKED ISSUE #5: ExtractAndExport: json serialization from odin?
I'd like to use Json serialization to get the me tions into a JSON file so that we can ingest it into solr. 

Is this code the write place to build on? https://github.com/clulab/processors/tree/master/main/src/main/scala/org/clulab/serialization/json

jpfairbanks:
Closes #5

### DIFF
--- a/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndExport.scala
+++ b/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndExport.scala
@@ -96,9 +96,6 @@ object SerializedMentions {
 
 
 object ExporterUtils {
-  // def tojson(a: Any) String {
-  //   write(a)
-  // }
 
   def removeTabAndNewline(s: String) = s.replaceAll("(\\n|\\t)", " ")
 }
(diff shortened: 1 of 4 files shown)


text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndExport.scala
```diff
@@ -64,11 +72,25 @@ case class SerializedExporter(filename: String) extends Exporter {
   override def close(): Unit = ()
 }
 
+case class JSONExporter(filename: String) extends Exporter {
+  override def export(mentions: Seq[Mention]): Unit = {
+    Serializer.save[SerializedMentions](new SerializedMentions(mentions), filename + ".json");
```
marcovzla: Let's delete this line because it is unnecessary for json serialization. It is actually using java serialization to write the mentions to the `.json` file, which is later overwritten with the actual json content (line 82).

jpfairbanks:
Thanks, this took me an silly amount of time because I had to lrn2scala while making the change. I went through a few iterations with a lot more code before I found the 
`import org.clulab.odin.serialization.json.JSONSerializer` object.

marcovzla:
Thank you @jpfairbanks!
Merging

PR #13, opened 2019-01-13
Data loader added for loading documents generated with science parse

bsharpataz:
added in the ExtractAndExport app, made a few things in ScienceParseDocument optional bc they were sometimes missing, perhaps more things should be Options...?
Also -- some design choices were made here, but can be revisited, primarily:
 - give grobid whole doc at a time, but any found grobid quantity that spans more than 1 sentence is ignored

Perhaps I should also add one that loads a plain text file and passes it to science parse in the moment... as an option? IDK, we'll prob be running science parse offline for our usage...

fyi @maxaalexeeva @jpfairbanks


text_reading/docker/docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
         image: allenai/scienceparse:2.0.3
         ports:
             - "8080:8080"
-    quantities:
+    grobid:
```
marcovzla: i wonder if we should name it `grobidquantities` instead, because grobid is a pdf parser that we're not using

text_reading/src/main/scala/org/clulab/aske/automates/scienceparse/ScienceParseClient.scala
```diff
@@ -74,7 +76,7 @@ class ScienceParseClient(
   }
 
   def parsePdf(file: File): ScienceParseDocument = {
-    val response = requests.post(url, headers = headers, data = file)
+    val response = requests.post(url, headers = headers, data = file, readTimeout = timeout, connectTimeout = timeout)
```
marcovzla: should we add this timeout to the other `requests.post` calls?

PR #14, opened 2019-01-19
Expand arxiv

bsharpataz:
Added script to expand arxiv
To note:
 - script is py2
 - flag keepall if set to `True` keeps the tarballs and pdfs from the uncompression, if left at false, these are removed (and I don't think we need them anyway)
- not parallelized, we/you may want to add that...?


bsharpataz:
@marcovzla todos addressed

bsharpataz:
Oh, and here's the "evidence" of the renaming :)
image

PR #21, opened 2019-01-28
Definition tests

maxaalexeeva:
@bsharpataz merged your tests with mine, added more tests, and edited mine for consistency. If there are three people working on the same files in different branches, should I be PR'ing every time I finish work for the day? Or do we just merge each other's branches?


bsharpataz:
either is fine -- whatever makes most sense to you!

PR #25, opened 2019-01-30
Masha tests

maxaalexeeva:
- Variable and Definition tests 
- Some minor changes to rules

Some rules and tests are commented out because they have been written but they do not produce the desired output and might need to be reconsidered/rewritten.


bsharpataz:
thanks!

PR #35, opened 2019-02-03
handle `include`

bsharpataz:
now handling the include env in the same way as import

softened exceptions to printed warnings bc it was crashing (they weren't being caught) at runtime


bsharpataz:
wouldn't it be funny if I responded with 👎  instead?  I bet the thumbs up and down have VERY diff frequencies in github PR threads... :)

marcovzla:
thanks!

PR #36, opened 2019-02-05
Becky tests

bsharpataz:
@zupon fyi
@maxaalexeeva please check over and let me know if you think this is a good idea -- we can use all, some, or none of it :)

### DIFF
--- a/text_reading/src/main/resources/automates.conf
+++ b/text_reading/src/main/resources/automates.conf
@@ -2,6 +2,7 @@ OdinEngine {
   // Override the default values here
 //  basePath = /org/clulab/aske_automates
 //  masterRulesPath = ${OdinEngine.basePath}/grammars/master.yml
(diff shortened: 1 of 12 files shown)


maxaalexeeva:
@bsharpataz, I looked through your changes and I think it all looks great!. I was wondering if the three of us could meet to go though the whole pipeline? (Thursday afternoon or we can set up something for next week). 
When we meet, I will also have questions on how to use the methods you wrote. E.g., def selectShorterAsVariable - can we use that to choose one of the two variable candidates (as opposed to applying it to **all** the mentions found) and if yes, how we can do that/where in the code that would go/ what the syntax would be, etc. 

Thank you!

bsharpataz:
@maxaalexeeva sure!  ok, I'll go ahead and merge it, and we can go through it, I'll send an email off-thread

PR #38, opened 2019-02-12
Regex spec char replacement

maxaalexeeva:
@bsharpataz, the only important changes are those in StringMatchEntityFinder.scala. I tried to fix that parsing error we had:
```
[info] - should extract variables from t1b: In APSIM, water uptake (Ta, mm d−1) is determined from potential transpiration demand (Tp, mm d−1), soil water available (WA, mm d−1), and water supply (WS, mm d−1) for each ith day and soil layer as: *** FAILED ***
[info]   org.clulab.odin.impl.OdinNamedCompileException: Error parsing rule 'stringmatch': Unmatched closing ')' near index 9
[info] Ta, mm d−1) is determined from potential transpiration demand (Tp, mm d−1
[info]          ^
[info]   at org.clulab.odin.impl.RuleReader.mkExtractor(RuleReader.scala:310)
[info]   at org.clulab.odin.impl.RuleReader.$anonfun$mkExtractors$4(RuleReader.scala:85)
[info]   at scala.collection.immutable.Stream.map(Stream.scala:415)
[info]   at org.clulab.odin.impl.RuleReader.mkExtractors(RuleReader.scala:85)
[info]   at org.clulab.odin.impl.RuleReader.readSimpleFile(RuleReader.scala:41)
[info]   at org.clulab.odin.impl.RuleReader.read(RuleReader.scala:31)
[info]   at org.clulab.odin.ExtractorEngine$.apply(ExtractorEngine.scala:87)
[info]   at org.clulab.aske.automates.entities.StringMatchEntityFinder.$anonfun$extract$1(StringMatchEntityFinder.scala:21)
[info]   at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
[info]   at scala.collection.immutable.HashSet$HashSet1.foreach(HashSet.scala:320)
[info]   ...

```

With the method I wrote, the error is not there anymore (and one more test got fixed), BUT a) I am not completely sure why that regex in line 53 worked/if the whole thing works correctly, b) I think more special symbols will need to be included c) it does not do what we originally planned to do (replace special symbols with _escape character + the special symbol_) --- I tried a bunch of different ways to make that to work but either the syntax was wrong or the output was not what we would want). Can we look at this together next week for a few minutes? 

The rest of the changes are just changes for consistency that I made last week and did not push; no changes in functionality or anything like that.

### DIFF
--- a/text_reading/src/main/scala/org/clulab/aske/automates/entities/StringMatchEntityFinder.scala
+++ b/text_reading/src/main/scala/org/clulab/aske/automates/entities/StringMatchEntityFinder.scala
@@ -15,7 +15,7 @@ class StringMatchEntityFinder(strings: Set[String], label: String) extends Entit
            |   priority: 1
            |   type: token
            |   pattern: |
-           |       /${stringToMatch}/
+           |       /\\Q${stringToMatch}\\E/
            |
         """.stripMargin
       engine = ExtractorEngine(ruleTemplate)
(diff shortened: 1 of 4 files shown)


maxaalexeeva:
@bsharpataz made the change we discussed. Only worked with double-slashes (java way to escape, I think).

bsharpataz:
thanks!

PR #43, opened 2019-02-19
Listc

maxaalexeeva:
This is supposed to handle type "listc" (a very very temporary fix), which is apparently a conjunctive list (https://github.com/kermitt2/grobid-quantities/blob/56bf7e13c48225539b07c5322a68cdd48e097a7a/src/main/java/org/grobid/core/utilities/UnitUtilities.java#L117)

If something is type listc, whatever is inside it is dealt with as atomic values (not intervals). This does not handle type listc as interval because grobid listc/interval output is not consistent. 

The before (a toy example):

image

The after:

image


bsharpataz:
Thanks!!  LGTM as far as I can tell

Have we ever seen the other one -- `listd` ?

maxaalexeeva:
@bsharpataz I haven't seen listd yet. I tried a few examples with OR instead of AND (listd is supposed to be a disjunctive list), but they still showed up as listc. I guess we'll know when we get a key listd not found error.

bsharpataz:
hahaha 👍

PR #46, opened 2019-02-22
Docker

bsharpataz:
poor mans parallelization and added the python files into the docker container to be able to run from any dir


marcovzla:
Thanks!

PR #49, opened 2019-02-25
Masha tests

maxaalexeeva:
-changed some rules to fix tests
-edited some tests
-changed failingTests back to passingTests (easier to track non-fixed tests this way)
-added some constraints to the expansion handler (when we are looking for variables, the longest among overlapping mentions is not the best choice... should we disable the expansions for the time being?)
-disabled intervals from grobid (they are now just returned as lists of values)
-turned "Allow exception to close parens" back on
-some other minor edits

### DIFF
--- a/demo/templates/about.html
+++ b/demo/templates/about.html
@@ -1,4 +1,4 @@
-      <div class="col-lg-3" style="padding-right: 3rem;">
+      <div class="col-lg-2" style="padding-right: 3rem;">
           Input Fortran code in the code editor area to the right and hit the
           'Submit' button. Two examples are provided to get started:
           <ul>
--- a/demo/templates/index.html
+++ b/demo/templates/index.html
@@ -8,7 +8,7 @@
         </div>
       <div class="row justify-content-md-center" style="padding-top: 1.5rem; padding-left:3rem; padding-right: 4rem;">
       {% include "about.html" %}
-      <div class="col-lg-3">
+      <div class="col-lg-4">
 		<div class="dropdown">
 		  <button class="btn btn-secondary btn-sm dropdown-toggle" type="button"
 				  id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true"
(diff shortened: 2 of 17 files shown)


maxaalexeeva:
@bsharpataz, would you have a look at this PR (the commits should be reasonably informative?) and tell me which of the stuff you are okay with? If you don't want to keep all of the commits, I can cherry pick the ones you agree with and make a new PR. Let me know if the commit messages are not informative enough and I can write more detailed explanations.

maxaalexeeva:
@bsharpataz, I realized that I didn't finish writing the method to test anything other than length the other day and I haven't figured out yet how to get it to test anything valuable yet:

https://github.com/ml4ai/automates/blob/6b60a42f150e228a9bb0dbc9d7985d40099e090a/text_reading/src/test/scala/org/clulab/aske/automates/TestUtils.scala#L73

I'll keep trying (unless you think we shouldn't make a separate label/test set for intervals).

bsharpataz:
> @bsharpataz, I realized that I didn't finish writing the method to test anything other than length the other day and I haven't figured out yet how to get it to test anything valuable yet:
> 
> [automates/text_reading/src/test/scala/org/clulab/aske/automates/TestUtils.scala](https://github.com/ml4ai/automates/blob/6b60a42f150e228a9bb0dbc9d7985d40099e090a/text_reading/src/test/scala/org/clulab/aske/automates/TestUtils.scala#L73)
> Line 73 in [6b60a42](/ml4ai/automates/commit/6b60a42f150e228a9bb0dbc9d7985d40099e090a)
>  def testThreeArgEvent(mentions: Seq[Mention], eventType: String, arg1Role: String, arg2Role: String, arg3Role: String, desired: Seq[Seq[String]]): Unit = { 
> I'll keep trying (unless you think we shouldn't make a separate label/test set for intervals).

Let's hold tight on intervals for now, for sure revisit, but let's shift focus.  We should meet soon to discuss, thanks!!

maxaalexeeva:
@BeckySharp, made the change (marked all the currently failing tests)

PR #55, opened 2019-03-18
Comment tests

maxaalexeeva:
- went back on this change from the previous PR because it looks like that broke several tests:
https://github.com/ml4ai/automates/blob/4214b0dbf8078a4fbb78f9c38a3142b51bc201d8/text_reading/src/main/scala/org/clulab/aske/automates/OdinEngine.scala#L79
- made changes to have comment tests use CommentEngine and not TestEngine;
- wrote a rule to extract vars and definitions from comments---the var tests pass, but not definition ones although the mentions as printed out from within the comment definition tests seem to look right; haven't figured out why the tests fail yet (upd: found the error in the rule that caused the tests to break);
- changed the comment tests not to include 'is' (we don't insert 'is' in preprocessing, so neither the rules, nor the tests will contain it);
- added a todo that forgot to add a while ago.

### DIFF
--- a/text_reading/src/main/resources/org/clulab/aske_automates/grammars/comments/definitions.yml
+++ b/text_reading/src/main/resources/org/clulab/aske_automates/grammars/comments/definitions.yml
@@ -8,50 +8,8 @@ rules:
     label: Definition
     priority: ${priority}
     type: token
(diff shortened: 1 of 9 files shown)


BeckySharp:
thanks!

PR #58, opened 2019-03-24
Fixing defs

zupon:
changes Masha and I made in the past ~2 days


maxaalexeeva:
@BeckySharp, the updates to the rules are minor and should not influence most recent changes in the entrypoint you made in a negative way, so I feel it should be safe for me to merge this; I need to merge this now, so that we could merge entrypoint into our branch without overwriting any changes we made.

PR #59, opened 2019-03-27
Finding param settings (intervals and more/less than)

maxaalexeeva:
added a rule to find intervals; 
added an avoid to help find intervals; 
edited rules for more_than and less_than (the token versions are still there for my reference);
added a rule for mislabeled adj for future use(commented out for now)


BeckySharp:
thanks!

PR #76, opened 2019-07-25
added the SIR-SEIR pdf, latex, and json

maxaalexeeva:


adarshp:
@maxaalexeeva Could you please `git rm` the generated PDF file? As a general rule of thumb let's try to keep automatically generated files (especially non-text files, since they don't play well with git) out of version control, since they will bloat the size of the repository.

Also, @marcovzla , after @maxaalexeeva removes the generated PDF file, could you please do a 'Squash and Merge' instead of a regular merge?

maxaalexeeva:
Thanks for letting me know!  I will be able to do that between 10 and 11 tonight. Sorry about the delay!

adarshp:
> Thanks for letting me know! I will be able to do that between 10 and 11 tonight. Sorry about the delay!

No worries, thanks @maxaalexeeva !

PR #80, opened 2019-08-09
moved syntax up a bit in webapp

BeckySharp:
Now just under the brat extraction viz
image


BeckySharp:
a better image example :)
image

maxaalexeeva:
@BeckySharp Yay! Much better! Thanks! And I looked at what you did to make that happen and I could swear I have tried that, but it wouldn't work ;(

BeckySharp:
ha, well, it took me a while to decide where to insert changes -- marco set most of this up so I'm still learning much of the ropes :)

PR #83, opened 2019-08-15
Reading comments from json

maxaalexeeva:
@BeckySharp, I have added the code to read the comments from grfn instead of txt, but let's merge your PR first and then I'll resolve whatever conflicts arise? (and whatever other comments you may have for this)

### DIFF
--- a/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
+++ b/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
@@ -14,6 +14,7 @@ import org.clulab.processors.Document
 import org.clulab.processors.fastnlp.FastNLPProcessor
 import org.clulab.utils.{DisplayUtils, FileUtils}
 import org.slf4j.LoggerFactory
+import ujson.Obj
 import upickle.default._
 
 import scala.collection.mutable.ArrayBuffer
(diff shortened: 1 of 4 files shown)


text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
```diff
@@ -123,6 +102,32 @@ object ExtractAndAlign {
...
+      for (cc <- commentComponents) if (commentObject.contains(cc)) {
+        val text = commentObject(cc).arr.map(_.str).mkString("")
+        if (text.length > 0) {
+          commentTexts.append(text)
+        }
+      }
+    }
+
+    // Parse the comment texts
+    val docs = commentTexts.map(parseCommentText(_, filename = Some(grfn("source").arr.head.str)))
```
BeckySharp: Since you're adding the filename here, let's use the doc.id down below where we make the link_elements.
Also -- should we add more context? i.e., should we append in the "head/neck/foot" and container name?

BeckySharp: the "down below is either line 187 or 195, IDK which number is your new one :)

BeckySharp:
If we have the container info available (with your Obj you created for comments) -- should we use that (with the `source` too) to make a better "source" in the LinkElement?

maxaalexeeva:
@BeckySharp, in the sourceLinkElement, I am pretty sure that info is in the variable itself: `@variable::SIR-Gillespie-SD::main::S0::0` if that's what you mean.

BeckySharp:
no -- I mean in the comment link element

maxaalexeeva:
Do you mean this source is not what you'd like to see, it's not where you'd rather it be, or that I should change the way I get to it?
```
"element_1": {
    "type":"comment_span",
    "source":"SIR-Gillespie-SD.f; @container::SIR-Gillespie-SD::@global::gillespie; head",
    "content":"n_R      current number of recovered members",
    "content_type":"null"
}
```

*I can also stop by if that's easier

BeckySharp:
ha, sorry - I didn't think that was what was printing, maybe I looked at an
old commit hash?
no that looks great!

maxaalexeeva:
Cool! I'll see if there are any conflicts and tag you for final review.

PR #86, opened 2019-10-09
grammar updates

maxaalexeeva:
Updates to conf and a rule.
This seems to eliminate false positives introduced by fixing GrFNEntityFinder in this commit https://github.com/ml4ai/automates/commit/0be3b28b7ed7db4ee88ce4f4550f64f763b1cd37

I am still not sure why that fix resulted in false positives, but these changes seem to bring the extractions closer to what they were before + another variable is extracted correctly from the comments (a greek letter one).

### DIFF
--- a/text_reading/src/main/resources/application.conf
+++ b/text_reading/src/main/resources/application.conf
@@ -47,8 +47,8 @@ CommentEngine {
 
   entityFinder {
     enabled = true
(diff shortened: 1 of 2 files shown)


maxaalexeeva:
closing for now; will revisit if needed.

PR #102, opened 2020-01-17
Apis

BeckySharp:
@maxaalexeeva as I was trying to get this to work, stumbled across some things that I *think* I cleaned up, but please check the changes and let me know if I am breaking some functionality accidentally.  if it looks good, please merge

fyi @cl4yton


text_reading/src/main/scala/org/clulab/aske/automates/OdinActions.scala
```diff
@@ -124,28 +124,31 @@ class OdinActions(val taxonomy: Taxonomy, expansionHandler: Option[ExpansionHand
...
-        case rm: RelationMention => m.arguments.getOrElse("variable", Seq()).head
-        case em: EventMention => m.arguments.getOrElse("variable", Seq()).head
-        case _ => ???
-      }
-      if varMention.words.length == 1
-      word = varMention.words.head
-      if word.length <= 6
-      if (word.toLowerCase != word // mixed case or all UPPER
+    def passesFilters(v: Mention, isArg: Boolean): Boolean = {
+      if ((v matches "Variable") && isArg) return true
```
maxaalexeeva: @BeckySharp, the condition in line 128 does not work (a test broke) because I also need this check when the Variable label is actually assigned, e.g., the word 'for' is found as a Variable by the rule 'var_calculation' because the word 'calculation' is frequently preceded by a variable (e.g., _Kcbmin calculation is implemented by..._), then the action looksLikeAVariable is run, and I know that it's not really a Variable. Does that make sense? I got rid of the condition.

text_reading/src/main/scala/org/clulab/aske/automates/OdinActions.scala
```diff
@@ -124,28 +124,31 @@ class OdinActions(val taxonomy: Taxonomy, expansionHandler: Option[ExpansionHand
...
-      }
-      if varMention.words.length == 1
-      word = varMention.words.head
-      if word.length <= 6
-      if (word.toLowerCase != word // mixed case or all UPPER
+    def passesFilters(v: Mention, isArg: Boolean): Boolean = {
+      if (v.words.length != 1) return false
+      // Else, the variable candidate has length 1
+      val word = v.words.head
+      if (word.length > 6) return false
```
maxaalexeeva: @BeckySharp, added another condition from the previous version of the action.

text_reading/src/main/scala/org/clulab/aske/automates/OdinActions.scala
BeckySharp: ah, poop.
Ok, I'll modify the taxonomy and add an entry for Variables found with a Gazetteer, which i need to pass everytime...

maxaalexeeva: Oh, I see what you were going for with that...

BeckySharp:
ok @maxaalexeeva please check it out again, I think I fixed the issue, though I added some technical debt... :/
That said, if you're ok with this, i can then let ppl know the api is working

maxaalexeeva:
@BeckySharp, tests pass. I don't have any objections. 
Will this kind of variable need to be incorporated into the rules?

BeckySharp:
Nope, bc it's a child of Variable, so all the rules that operate over
Variables would accept these too.
Thanks for checking it out for me!  I'll go ahead and merge

maxaalexeeva:
Ah! I see! Sounds good!

PR #105, opened 2020-03-02
Access ontology

maxaalexeeva:


text_reading/src/main/scala/org/clulab/grounding/Grounder.scala
```diff
@@ -0,0 +1,222 @@
...
+      Map(mention.arguments("variable").head.text -> getTopK(allResults, k))
+    } else Map(mention.arguments("variable").head.text -> Array(new sparqlResult("None", "None", "None", None)))
+  }
+
+  //ground a string (for API)
+  def groundString(text:String): String = {
+    val terms = getTerms(text)
+    val resultsFromAllTerms = new ArrayBuffer[sparqlResult]()
+    for (word <- terms) {
+      val result = runSparqlQuery(word, "/home/alexeeva/Repos/automates/text_reading/sparql") //todo: pass from configs
```
BeckySharp: should we change this for this PR?

text_reading/src/main/scala/org/clulab/grounding/Grounder.scala
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  /* grounding one mention; return a map from the name if the variable from the mention to its svo grounding */
+  def groundOneMentionWithSparql(mention: Mention, k: Int): Map[String, Seq[sparqlResult]] = {
+
+    val terms = getTerms(mention) //get terms gets nouns, verbs, and adjs, and also returns reasonable collocations (multi-word combinations), e.g., syntactic head of the mention + (>compound | >amod)
+    if (terms.nonEmpty) {
+      val resultsFromAllTerms = new ArrayBuffer[sparqlResult]()
+      for (word <- terms.get) {
+        val result = runSparqlQuery(word, "/home/alexeeva/Repos/automates/text_reading/sparql") //todo: pass through configs
```
BeckySharp: should we pass this for this PR?

text_reading/src/main/scala/org/clulab/grounding/svoOntology.yml
```diff
@@ -0,0 +1,8 @@
+- phenomenon:
+    - spatiotemporal_phenomenon:
```
BeckySharp: what is this used for?

maxaalexeeva: Not used. Forgot to delete. Thanks!

text_reading/src/main/scala/org/clulab/grounding/Grounder.scala
maxaalexeeva: changed this to passing from configs, but it's stil an absolute path :/ will look into that again at some point tomorrow

maxaalexeeva: okay, I think it works now.

PR #118, opened 2020-03-26
TR updates

maxaalexeeva:
@BeckySharp, this pr sort of grew from the original name of the branch and includes the following things (some are to get better extractions, some are to include things I happen to need from time to time---basically some methods for my convenience if that's okay)

1. filtering out variables that consist of just punctuation---for now it's just a list; would you like this done in some other way?
2. an app to parse text (e.g., our science parse jsons) and produce processors Document (these are used, for example, for indexing to be used with taxero)
3. an exporter to export mentions to tsv, which I would like to have because I find tsv's more readable than jsons and tsv files are smaller than json files that we get with other methods => my laptop frequently takes forever to open the jsons we produce.
4. passing the path to the grounder from conf instead of getting the path based on current dir---that does not work after the repo re-arrangement.


src/text_reading/src/main/scala/org/clulab/aske/automates/OdinActions.scala
```diff
@@ -126,11 +126,12 @@ class OdinActions(val taxonomy: Taxonomy, expansionHandler: Option[ExpansionHand
 
   def looksLikeAVariable(mentions: Seq[Mention], state: State): Seq[Mention] = {
 
-
+    val punctuation = List(",", ".", "(", ")", "[", "]", "/", "~", "%")
     //returns mentions that look like a variable
     def passesFilters(v: Mention, isArg: Boolean): Boolean = {
       // If the variable was found with a Gazetteer passed through the webservice, keep it
       if ((v matches OdinEngine.VARIABLE_GAZETTEER_LABEL) && isArg) return true
+      if (v.words.length == 1 && punctuation.contains(v.words.head)) return false
```
BeckySharp: I think you could do this with the `isLetter` method, something like:
```
if (v.words.length == 1 && !(v.words.head.count(_.isLetter) > 0)) return false
```

then you wouldn't need a punct list

src/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExportDoc.scala
```diff
@@ -0,0 +1,72 @@
...
+  files.foreach { file =>
+    // 1. Open corresponding output file and make all desired exporters
+    println(s"Extracting from ${file.getName}")
+    // 2. Get the input file contents
+    //here, for the science parse texts, we join all the sections to produce the Document of the whole paper
+    val text = dataLoader.loadFile(file).mkString(" ")
+    val document = proc.annotate(text)
+    //jsonify the Document
+    val json = document.json(pretty = true)
+    //replace the file name with the name that indicates it has been serialized; todo: need a better way to indictae the target dir
```
BeckySharp: maybe add "_serialized.json" so you know it's serialized with json, not with java serialization

BeckySharp: Oh, I see, you add it in the JsonDocExporter, 👍

src/text_reading/src/main/scala/org/clulab/aske/automates/OdinActions.scala
maxaalexeeva: That's a great idea! Thanks, @BeckySharp! Didn't occur to me!

maxaalexeeva:
@BeckySharp, I've pushed the punctuation change. Thanks!

PR #120, opened 2020-03-27
added script to preprocess eqn imgs

BeckySharp:


BeckySharp:
FYI @cl4yton 

To run it you need an env with:
 - pillow
 - pypng
 - numpy

PR #121, opened 2020-03-27
Endpoint modification

BeckySharp:
@pauldhein please check this out and verify that it does what you were hoping for
The `mentions` key now should correspond to the filepath for the mentions file, which I assume you saved.  The Mentions in there should be json.

FYI @maxaalexeeva


pauldhein:
@BeckySharp it looks like I am able to ping the pipeline just fine now with the three JSON file paths. However, I am now getting the following error:
Screen Shot 2020-03-27 at 2 40 17 PM

It looks like I am missing `vectors.txt`. Do you know what that file is and where it is supposed to be? I can't find it in my local checkout of `automates`.

maxaalexeeva:
@pauldhein, vectors.txt file should be some sort of word vectors file. Mine, if I remember correctly, are glove vectors with a header line added (looks like this:
Screenshot from 2020-03-27 12-24-53

I had the file in text_reading dir, but there are probably better places for it. I would assume that if you had any text reading components running before, you probably had that file on you machine somewhere?

adarshp:
@pauldhein @maxaalexeeva Can we have vectors.txt automatically downloaded if it's not there? Keith implemented this with Eidos and the Glove embeddings in https://github.com/clulab/eidos/pull/746 

image

adarshp:
Not super urgent, but would be good for the future I think

BeckySharp:
Yeah, it won't be in a checkout -- it's the file with word embeddings.  Is the glove file somewhere in the google drive?That's really the set to use, I use vectors.txt locally to debug faster (smaller file), and guess I accidentally committed that.

BeckySharp:
> @pauldhein @maxaalexeeva Can we have vectors.txt automatically downloaded if it's not there? Keith implemented this with Eidos and the Glove embeddings in [clulab/eidos#746](https://github.com/clulab/eidos/pull/746)
> 
> image

I think thats a good solution, and if you keep the "path" the same (i.e., the same jar from (maven?)), then I won't need 2 copies in my `.ivy2` folder....

BeckySharp:
and to be clear -- I wasn't volunteering to do this... :)

maxaalexeeva:
@BeckySharp, as a temp solution, @pauldhein now has the link to the glove vectors Marco shared via google drive for taxero the other day and also my smaller vectors.txt file.

pauldhein:
@adarshp that is a great suggestion for how we can download large files needed for TR/ER after a repo checkout. We should definitely discuss this at a later date. I have a few other options in mind, such as a single folder in the AutoMATES google drive that we require DARPA to download, or a composed docker container that handles all downloading of large files necessary.

@BeckySharp and @maxaalexeeva thank you for your help in getting me some test vectors files! I did indeed have some on my machine as well but didn't have my path set appropriately. I'm continuing to test the `align` endpoint now and will update with future bugs in the `automates_text` slack channel.

PR #124, opened 2020-03-29
Grounder 2.0

maxaalexeeva:
@pauldhein, the `groundMentionToSVO` endpoint seems to work on my machine. It's still slow--- @BeckySharp, if you somehow find yourself with nothing to do, would you check the endpoint out?---at least for the reason that sparql queries are not instant (you can test it out here http://www.geoscienceontology.org/sparql.html --- sometimes it's fast, sometimes not so much) and also because there are 122 mentions to ground. I am not writing any of them to file---you can do that on your end, right? If you can write them to file, you should probably be able to read them in the align here https://github.com/ml4ai/automates/blob/0e9cf44554980e067cc0d1d74edae8adef4b0929/src/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala#L64 and uncomment svo grounder lines everywhere and then it should produce link hypothses for svo groundings. Let me know if that sounds ok.


maxaalexeeva:
@pauldhein, it actually ran through. Took roughly half an hour I think.

PR #125, opened 2020-03-30
deserializing svo groundings 

maxaalexeeva:
...and making link hypotheses for them in align.

@pauldhein, this is how you would include previously extracted and serialized svo groundings. it looks like it worked on my laptop. let me know if that works for you. you will need to update the path to svo gr json in tr_integration.

If it works, I will clean up the code tomorrow (I had to comment out the main method in extract and align to test this quickly and might want a separate method for reading in / deserializing the groundings)

### DIFF
--- a/scripts/model_analysis/tr_integration.py
+++ b/scripts/model_analysis/tr_integration.py
@@ -35,6 +35,7 @@ def test_align():
         f"{webservice}/align",
         headers={"Content-type": "application/json"},
         json={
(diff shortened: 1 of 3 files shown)


maxaalexeeva:
@pauldhein, for this to work, you will need to pass the svo_groundings.json to the extract_align_ground.py script, e.g., like so (screenshot from when it was tr_integration.py):
Screenshot from 2020-03-30 11-32-34

PR #136, opened 2020-04-10
Svo terms in text link el

maxaalexeeva:
Hi @pauldhein,

The changes in this commit https://github.com/ml4ai/automates/commit/d0c68b39a49e8deac5887bd01369f0f6383ed392 should fix your svo error from yesterday (the error was because there was a regex special character in the query term---the closing paren). 

For this commit, I was working on the same local branch that I used for my previous PR---big mistake. For some reason, the commit that should have already been merged (and I'm pretty sure the changes are already in endpoint) shows up here as well. Do you know what that's about? Hope that does not break anything. 

@BeckySharp, just fyi---that commit is what I ended up doing at least as a temp solution for the special character in queries issue we discussed.


maxaalexeeva:
Looks like my mistake with the branches is causing the checks to fail, so I am gonna close this PR and do it the proper way. @pauldhein @BeckySharp fyi

PR #138, opened 2020-04-12
Separating align

maxaalexeeva:
@BeckySharp, this is a proposal on how to separate align into different endpoints. This is the gist:

1. Using options in various align methods (e.g., in getLinkElements and mkLinkHypothesis) to avoid breaking up linkElement and linkHypothesis-related methods---there is finally a full pipeline, and I hate to break it up. 

2. For every align-related endpoint (see _Note_),  there is a method that gets all the necessary arguments depending on what we want to link:
https://github.com/ml4ai/automates/blob/1367c4a0fd65b34c0ce5688235eb5f8c3303acf9/src/text_reading/webapp/app/controllers/HomeController.scala#L249
There's a _toAlign_ list in each endpoint, telling it what needs aligned, e.g., https://github.com/ml4ai/automates/blob/1367c4a0fd65b34c0ce5688235eb5f8c3303acf9/src/text_reading/webapp/app/controllers/HomeController.scala#L192

_Note:_ now, as a prototype, there are two endpoints: 
- align---to align everything
- alignDocstringsAndText---that does not include equation linking, with equation production was a bottleneck that stopped us from running the pipeline 

3. Fwiw, it's usable for Paul to run the pipeline for cases when some of the input, e.g., equations, is missing, but we can definitely discuss a better way to do this in our Monday meeting.  

Also, `main` method in `ExtractAndAlign` is temporarily commented out---I will adjust it after the other changes are finalized.

### DIFF
--- a/src/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
+++ b/src/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
@@ -22,6 +22,7 @@ import org.json4s
 
 import scala.collection.mutable.ArrayBuffer
 
+case class alignmentArguments(grfn: Value, variableNames: Seq[String], variableShortNames: Seq[String], commentDefinitionMentions: Option[Seq[Mention]], definitionMentions: Option[Seq[Mention]], equationChunksAndSource: Option[Seq[(String, String)]])
 
 object ExtractAndAlign {
   val COMMENT = "comment"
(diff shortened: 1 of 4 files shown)


pauldhein:
## Proposed new JSON schema
The JSON that will be passed to the `/align` endpoint will be an object with keys that can be used to determine which parts of the alignment are supposed to be carried out.

### Alignment keys
I plan on passing the following keys to `/align` ... @maxaalexeeva please let me know if these will work for you:
- `mentions`: the mentions object that I get from the `/pdf_to_mentions` endpoint
- `equations`: an array strings where each string is a tokenized LaTeX equation
- `source_code`: a JSON object its own keys `variables` and `comments`
  - `variables`: the list of variables as you are used to seeing from the `grfn` object
  - `comments`: the extracted source code comments as you currently get from the `grfn` object (currently named `source_comments` but the prefix is no longer necessary)

### Which alignments to perform
Below is my impression of which alignments will be performed based on the keys I pass in to `/align`. @maxaalexeeva can you confirm that the endpoint will behave as I'm expecting?
```python
if keys == ["mentions", "equations", "source_code"]:
    # align text mentions and equations
    # align source variables and source comments
    # align text mentions and source comments
else if keys == ["mentions", "equations"]:
    # align text mentions and equations
else if keys == ["mentions", "source_code"]:
    # align source variables and source comments
    # align text mentions and source comments
else:
    # output a warning: "Nothing to do for keys: {(place the set of keys here)}"
```

### Alignment results
As we discussed in the meeting with @BeckySharp the results from `/align` should be a JSON array of link hypotheses. @maxaalexeeva please let me know if this is not the case.

maxaalexeeva:
@BeckySharp @pauldhein, this branch is mainly done. Some highlights:

- GrFNParser is mainly untouched and that's what the main method in extract and align works with; some methods from grfnparser that have to be slightly different for the way they are in grfnParser to accommodate the new json input are in https://github.com/ml4ai/automates/blob/68502d7cb73738777bf64c12c2d89aca6700412d/src/text_reading/src/main/scala/org/clulab/utils/AlignmentUtils.scala

- there is a new package that converts ujson to json4s (to rehydrate the mentions)---it runs, but intellij, for some reason, marks methods using that conversion red... Does anyone else have that issue?

- there have been some modifications to a couple of methods to accommodate both json and grfn (eg in https://github.com/ml4ai/automates/blob/68502d7cb73738777bf64c12c2d89aca6700412d/src/text_reading/src/main/scala/org/clulab/aske/automates/entities/StringMatchEntityFinder.scala)

- there are options all over the place

- I pulled master, so some files modified are not mine (e.g., https://github.com/ml4ai/automates/blob/68502d7cb73738777bf64c12c2d89aca6700412d/scripts/text_reading/align_json_package.py)

I am working on fixing the equation-text_span linking issue on a different branch.

maxaalexeeva:
@pauldhein, again, the branch is not cleaned up, but it is ready for you to check if the functionality looks right. For grounding to svo within align, there are 3 options (I will document this in the wiki when we have finalized align):

1. run align without svo grounding; should be set to false here: https://github.com/ml4ai/automates/blob/4d8b27cf42b333d5456ab7c54c988cf59307f23a/src/text_reading/webapp/app/controllers/HomeController.scala#L191 (I am accepting suggestions on where to put this toggle)

2. ground to svo within align (somewhat slow, but a lot of redundancy has been eliminated) (groundToSVO=true)

3. use svo groundings you get from the groundToSVO endpoint (`groundMentionsToSVO`); for that, add these groundings to the input json with the key "SVOgroundings" 

Also, the result of align can now be either a json object or write to the input file like it did before (https://github.com/ml4ai/automates/blob/4d8b27cf42b333d5456ab7c54c988cf59307f23a/src/text_reading/webapp/app/controllers/HomeController.scala#L192)

All of these seem to run fine on my machine. When you have tested them, I will do some more cleanup (print statements and such). There are some comments to non-transparent methods, but let me know if you need me to clarify some things (either in writing or via zoom).

maxaalexeeva:
@pauldhein, also, I had to update some dependencies.

maxaalexeeva:
Hi, @pauldhein:

Now the input json (the one with mentions, svo groundings, etc), can include the toggles and args. You'd just add this to the json object:

```
"toggles": {"groundToSVO": true, "appendToGrFN": false},
  "arguments": {"maxSVOgroundingsPerVar": 5},
```

If these are not in the input, it will default to whatever is set in the HomeController class (e.g., in case we run an old GrFN, which does not have these fields).

PR #159, opened 2020-10-07
Equation MathJax scripts: refactoring and adding MathML rendering helper

cl4yton:
In this pull request:

- Removes: render-v0, render-v1, server (this latter "server" was the flask-based one you showed me); I'm still keeping all three locally as I program by example, but I think they're not needed in the repo
- Renaming:
  - webservice -> webservice_convert_to_mathml (I know this is verbose, but I'd like to it be clear what each of these things does)
  - render -> render_latex+mathml
- Adds new script will just be render_mathml : renders one (or optionally 2) columns of MathML (2-columns for comparison), showing source and MathJax-rendered equations.

@pauldhein : you can merge and delete this branch once reviewed.


pauldhein (approved):
The additions look good, and all files that aren't supposed to be in the repo are also still present in master, so I see no reason to wait on merging these changes.

PR #168, opened 2020-12-09
improving grammars

maxaalexeeva:


automates/text_reading/src/main/resources/application.conf
```diff
@@ -60,7 +60,8 @@ CommentEngine {
 
 apps {
 
-  projectDir = "/home/marco/repos/automates"
+projectDir = "/home/alexeeva/Repos/automates"
```
pauldhein, line 63: @maxaalexeeva It would be great if we set this variable to a non-specific user default that would require all users to override the variable when starting a TextReading web app session. Maybe something like `projectDir = "/local/path/to/automates"`

Can we do that? Do you remember the command that allows a user to override an environment variable when running `sbt`?

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/definitions.yml
```diff
@@ -51,7 +51,20 @@ rules:
...
-      #the !b-variable helps find greek letters as coefficients by avoiding finding variables preceeded by greek letters
+#      |
+#      (?<definition> [word = /.*/]) (?<variable> [entity = "B-GreekLetter"])
+
+
+  - name: var_stands_for_def
+    label: Definition
+    priority: ${priority}
+    type: dependency
+    example: "V stands for vendetta."
```
pauldhein, line 62: @maxaalexeeva  I love the reference!

automates/text_reading/src/main/scala/org/clulab/aske/automates/OdinActions.scala
```diff
@@ -133,7 +142,7 @@ class OdinActions(val taxonomy: Taxonomy, expansionHandler: Option[ExpansionHand
       // If the variable was found with a Gazetteer passed through the webservice, keep it
       if ((v matches OdinEngine.VARIABLE_GAZETTEER_LABEL) && isArg) return true
       if (v.words.length == 1 && !(v.words.head.count(_.isLetter) > 0)) return false
-      if (v.words.length > 1 && v.entities.get.exists(m => m matches "B-GreekLetter")) return true //account for var that include a greek letter---those are found as separate words even if there is not space
+      if ((v.words.length > 1 || v.words.length == 1) && v.entities.get.exists(m => m matches "B-GreekLetter")) return true //account for var that include a greek letter---those are found as separate words even if there is not space
```
pauldhein, line 145: @maxaalexeeva please change this to `v.words.length >= 1`

pauldhein (changes requested):
@maxaalexeeva  I only have two minor change requests and then I'll merge this PR.

PR #171, opened 2021-01-15
Update GrFN from json functions

dpdicken:
* Update GrFN from json functions to correctly build the graph
* Update how function sets are computed per function/container
* Rename PASS nodes to INTERFACE
* Update var type fields to data_type


automates/model_assembly/networks.py
```diff
@@ -270,18 +272,21 @@ def __eq__(self, other) -> bool:
...
             and self.border_color == other.border_color
-            and set([n.uid for n in self.nodes])
-            == set([n.uid for n in other.nodes])
+            and set([n.uid for n in self.nodes]) == set([n.uid for n in other.nodes])
         )
 
     @classmethod
-    def from_container(cls, con: GenericContainer, occ: int):
+    def from_container(
+        cls, con: GenericContainer, occ: int, parent_subgraph: "GrFNSubgraph"
```
pauldhein, line 280: @dpdicken why is the type annotation for `parent_subgraph` the string `"GrFNSubgraph"`? Is there some reason why we can't use the class name itself like we did for `con`?

automates/model_assembly/sandbox.py
```diff
@@ -38,8 +38,8 @@ def load_lambda_function(func_str: str) -> Callable:
         raise UnsafeOperationError(f"found in lambda:\n{func_str}")
 
     # Checking for expected lambda expression header
-    if re.match(r"lambda [A-Za-z0-9_,]*:", func_str) is None:
-        raise RuntimeError(f"Unexpected form for lambda:\n{func_str}")
+    # if re.match(r"lambda [A-Za-z0-9_,]*:", func_str) is None:
```
pauldhein, line 41: @dpdicken why did you comment out this check? I use this to ensure we don't see any code injection or malformed members amongst the arguments of a lambda function.

pauldhein (commented):
Thanks for the great changes @dpdicken! I have only two comments for you to address in this review.

pauldhein:
Oh @dpdicken please check the issue found with CodeFactor as well. That is our new automated code review bot that will be helping me review PRs. From the issue details I found that the following line of code in `networks.py::630`:
```python
if type(n) == LambdaNode
```
... not sure if this is your line or one of mine from long ago, but you'll want to change this line to:
```python
if isinstance(n, LambdaNode)
```

automates/model_assembly/networks.py
dpdicken: When doing references to the class that its contained within, you have to put it as a string literal. Another option is to add `from __future__ import annotations` as an import and the module will automatically convert all type hints to strings for you so you dont have to manually do it in these situations.  They are supposed to handle this in python 3.10. 

(I got this info from here :) https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class)

automates/model_assembly/sandbox.py
dpdicken: Thanks for catching this, I had commented it out because a test was failing due to the lambda regex match here failing. It was failing even though the lambda looked perfectly fine to match with the regex to me. However, I just uncommented it and it looks like its passing fine now.

automates/model_assembly/networks.py
pauldhein: @dpdicken thanks for the pro-tip! Let's go ahead and use the future annotations import.

pauldhein (changes requested):
The changes included look good. @dpdicken can you please add the `from __future__ import annotations` change? After that I will approve and merge this branch.

automates/model_assembly/networks.py
pauldhein: I went ahead and did the update that should resolve this conversation.

pauldhein (approved):
Merge ready!

PR #177, opened 2021-01-28
Cosmos

maxaalexeeva:
- Added an endpoint to extract mentions from cosmos json (`call_cosmos_json_to_mentions()`)
- Added serializer to store mentions that contain attachments storing mention location information from cosmos
- Adjusted some methods in the `align` endpoint to use mentions extracted from cosmos and add "spans" info to align output 
- When running align with science parse, add an extra argument ("text_input_format") to "arguments" in payload, like this:

```
"arguments": {"maxSVOgroundingsPerVar": 5, "text_input_format": "science-parse"}
```

Otherwise, either use "cosmos" as the value or omit this argument. 

To ping endpoint:

```
def call_cosmos_json_to_mentions():
    res = requests.post(
        f"{webservice}/cosmos_json_to_mentions",
        headers={"Content-type": "application/json"},
        json={
            "pathToJson": "/home/alexeeva/Repos/automates/scripts/model_assembly/cosmos_to_mentions_payload.json"
        },
    )

    json_dict = res.json()
    json.dump(json_dict, open("masha8.json", "w"), indent=4)
```

Payload:

```
{
  "path_to_cosmos_json": "/media/alexeeva/ee9cacfc-30ac-4859-875f-728f0764925c/storage/automates-related/2003_cosmos_json.json",
  "debug": false
}
```

Sample align output for text var:

```
"text_var": [
{
            "uid": "9416264b-7a4d-4d7a-9407-d9e1ba78ae89",
            "source": "unk_text_file_sent11_Interval(19, 21)",
            "original_sentence": "Both epidemics would spread in parallel , and it can be expected that the epi - demic caused by virus B which is rather innocuous , pro - tects against SARS ( so that na i ve regions , not protected by the epidemic B can get SARS large outbreaks ) .",
            "identifier": "B",
            "definition": "virus",
            "svo_terms": "virus",
            "unit": "None",
            "paramSetting": {
                "value": "None",
                "valueLeast": "None",
                "valueMost": "None"
            },
            "svo_groundings": "N/A",
            "spans": [
                {
                    "page": 2,
                    "block": 1,
                    "span": {
                        "char_begin": 1867,
                        "char_end": 1874
                    }
                }
            ]
        }
]
```


maxaalexeeva:
@pauldhein , fyi, in the next pr, the `text_input_format` is switched to `serializerName` because that ended up a more crucial difference, but I don't think you need to worry about that - I set the default serializer to the one we need and I'm pretty sure it shouldn't matter on your end. I'll let you know if it turns out it does (or I will be able to identify the error if you get it).

automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
```diff
@@ -536,9 +543,35 @@ object ExtractAndAlign {
...
+            "page" -> page,
+            "block" -> block,
+            "span" -> ujson.Obj(
+              "char_begin" -> charBegin,
+              "char_end" -> charEnd
+                      )
+                    )
+        } else {
+          ujson.Obj(
+            "page" -> "N/A",
```
pauldhein, line 565: @maxaalexeeva please use `null` as opposed to the string _N/A_. That will produce the null object in JSON that will be read as `None` in Python. We are using `None` to be able to tell when a key does not have any object across all of our JSON specs in AutoMATES.

Additionally, I think you are outputting the string "None" in some locations. Can you please change those to be `null` as well?

automates/text_reading/webapp/app/controllers/HomeController.scala
```diff
@@ -219,10 +261,17 @@ class HomeController @Inject()(cc: ControllerComponents) extends AbstractControl
       json("arguments").obj("maxSVOgroundingsPerVar").num.toInt
     } else maxSVOgroundingsPerVarDefault
 
+    val textInputFormat = if (jsonKeys.contains("arguments")) {
+      val args = json("arguments")
+      if (args.obj.keys.toList.contains("text_input_format")) {
```
pauldhein, line 266: @maxaalexeeva you should be able to do the following for the key membership check on line 266:
```scala
args.obj.contains("text_input_format")
```

This method should attempt to hash the string `text_input_format` into the `args` object. That is an _O(1)_ operation, whereas (if I am understanding this correctly) then your method is checking if the string `text_input_format` is in the list of keys that are a part of the object and that is an _O(n)_ operation. Efficiency probably isn't too big of a deal here, but we should always strive for _O(1)_ operations when the opportunity presents itself.

Try out the method I have above and see if it works (I'm not actually sure if the syntax is correct). I found this [link](https://alvinalexander.com/scala/how-to-text-existence-key-value-map-scala-cookbook/) that seems to think we can use that syntax. If it doesn't work then I'll search for another way to do an _O(1)_ check.

automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
maxaalexeeva: @pauldhein good points. I have addressed the null issue to the extent that we do not output any none or NA strings in the align endpoint. There are still some "None"s here and there in the code, but none of them make it into the output. I will try to get rid of those as I adjust the endpoints based on the meeting Fri, Feb 18, but I'd like to not spend more time on this change right now since some of that code might be gone anyway. Does that work?

automates/text_reading/webapp/app/controllers/HomeController.scala
maxaalexeeva: @pauldhein, nice! Didn't know of that way. I think it worked. Also, I use that site you linked a lot! Great explanations!

automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
pauldhein: @maxaalexeeva as long as you addressed the N/As here then we can move forward with the merge. I agree that you shouldn't spend any more time tracking down additional N/As or nones until we know what code will be staying in the repo.

automates/text_reading/webapp/app/controllers/HomeController.scala
pauldhein: Glad it worked. Thank you for making the change!

pauldhein (approved):
Looks like all my requested changes are done and this branch is ready to merge!

PR #185, opened 2021-02-18
CAST to GrFN initial implementation

dpdicken:


automates/program_analysis/CAST2GrFN/model/cast_to_air_model.py
```diff
@@ -0,0 +1,417 @@
+from typing import List, Dict
+from enum import Enum
+
+from automates.program_analysis.CAST2GrFN.model.cast import AstNode
+
+
+class CASTToAIRException(Exception):
```
pauldhein, line 7: @dpdicken lets stick with the `obj2obj` naming scheme here and change the class name to: `C2A<error-type>`.

Additionally, I would like to break down this single exception into a family of exceptions. Minimally I would like the following to be included:
- `C2ATypeError`: inherits from TypeError and is used for any type errors that occur during CAST --> AIR processing (see `cast_to_air_visitor.py::101`)
- `C2ARuntimeError`: inherits from RuntimeError and is used for any runtime errors that occur during CAST --> AIR processing
- `C2ANameError`: inherits from NameError and is used when name errors occur (such as a missing member variable for some object) during CAST --> AIR processing
- `C2AValueError`: inherits from ValueError and is used when an operation cannot be performed for a given value during CAST --> AIR processing

automates/program_analysis/CAST2GrFN/model/cast_to_air_model.py
```diff
@@ -0,0 +1,417 @@
...
+
+
+class C2AIdentifierType(str, Enum):
+    UNKNOWN = "unknown"
+    VARIABLE = "variable"
+    CONTAINER = "container"
+    LAMBDA = "lambda"
+
+
+class C2AIdentifierInformation(object):
```
pauldhein, line 25: @dpdicken The definition of this class (and **many** subsequent classes in this file) is very verbose. Is there a reason we can't use Python's [dataclasses](https://docs.python.org/3/library/dataclasses.html) for these class definitions? Please see the definition of `XXIdentifier` classes in `model_assembly/structures.py` for an example of how to use inheritance and dataclasses together.

automates/program_analysis/CAST2GrFN/visitors/cast_visitor.py
```diff
@@ -0,0 +1,42 @@
...
+
+class CASTVisitor:
+    def __init__(self):
+        pass
+
+    def visit_list(self, node_list: List):
+        return [self.visit(n) for n in node_list]
+
+    @methdispatch
+    def visit(self, node: AstNode):
```
pauldhein, line 41: @dpdicken see my previous comment in `cast_to_air_visitor.py` ... this error message works for now, but once all expected `AstNode`s are handled please convert this to a `TypeError`.

automates/utils/method_dispatch.py
```diff
@@ -0,0 +1,12 @@
+from functools import singledispatch, update_wrapper
+
+# TODO Check if this can be removed now that we use python 3.8
+def methdispatch(func):
```
pauldhein, line 4: @dpdicken Can you explain to me what is going on here? Simple `@singledispatch` wrappers have worked fine for me for some time, even with multiple args and kwargs. The only limitation seems to be that dispatch can only operate on the first argument. Are you doing something else here? The Python [docs](https://docs.python.org/3.8/library/functools.html#functools.singledispatch) seem to suggest that additional args and kwargs will work just fine.

tests/program_analysis/CAST2GrFN/test_cast_json.py
```diff
@@ -0,0 +1,164 @@
...
+    UnaryOp,
+    UnaryOperator,
+    VarType,
+    Var,
+)
+
+from automates.program_analysis.CAST2GrFN.cast import CAST, CASTJsonException
+
+
+def get_file_str(filepath):
```
pauldhein, line 36: @dpdicken this design pattern is unnecessary for JSON files. Please use the following instead:

```python
import json

my_data = json.load(open("<filepath>", "r"))
```

tests/program_analysis/CAST2GrFN/test_cast_json.py
```diff
@@ -0,0 +1,164 @@
...
+def test_all_nodes_to_json_and_from_result(
+    cast_with_all_nodes_json, cast_with_all_nodes
+):
+    cast_json = cast_with_all_nodes.to_json_str()
+    assert cast_json == cast_with_all_nodes_json
+    cast_from_generated_json = CAST.from_json_str(cast_json)
+    assert cast_with_all_nodes == cast_from_generated_json
+
+
+def test_from_json_no_node_type_exception(cast_with_no_node_type_exception_json):
```
pauldhein, line 143: @dpdicken I like the ingenuity behind these error testing examples! However, `pytest` provides us with a cleaner way to conduct exception testing. Please use the design pattern below to remake this test and all other error/exception testing:

```python
import pytest

def test_zero_division():
    with pytest.raises(ZeroDivisionError):
        1 / 0
```

tests/program_analysis/CAST2GrFN/test_cast_to_grfn.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+def test_basic_function_def_and_assignment(
+    basic_function_def_and_assignment_grfn, basic_function_def_and_assignment_cast
+):
+    generated_grfn = basic_function_def_and_assignment_cast.to_GrFN()
+    assert generated_grfn == basic_function_def_and_assignment_grfn
+
+
+def test_cast_with_all_nodes(cast_with_all_nodes_grfn, cast_with_all_nodes):
```
pauldhein, line 329: @dpdicken using an `assert True` for a test case will artificially increase the total number of passing tests in our repo. We don't want that to happen for the purposes of gathering accurate stats about passing tests. Please convert tests that have not been fully implemented to the following form:
```python
@pytest.mark.skip("Test is unfinished")
def test_cast_with_all_nodes(cast_with_all_nodes_grfn, cast_with_all_nodes):
    return NotImplemented
```

pauldhein (changes requested):
@dpdicken first off, excellent job on getting this mountain of work ready so quickly!

I have a series of comments for you to address before this PR is ready to merge. Most of the comments stem from Python design patterns or specific design choices. To the extent that we need to discuss any of my comments lets keep the discussion in this PR thread (but please ping me on Slack when you have added comments).

automates/utils/method_dispatch.py
dpdicken: The issue is I wanted to use single dispatch on methods in a class in which their first argument is self! So, it breaks the normal use of single dispatch for class methods because you can't have variable types of self being passed in. So, this function allows you to use single dispatch on class methods by working around that problem. They actually implemented a version of this "methdispatch" in functors in python 3.8 to avoid this problem, which is why I have the todo there to check if that will work instead.

automates/program_analysis/CAST2GrFN/model/cast_to_air_model.py
dpdicken: sure, I don't see why we shouldn't use that.

automates/utils/method_dispatch.py
pauldhein: Cool, yeah I just found the [docs](https://docs.python.org/3/library/functools.html#functools.singledispatchmethod) for `@singledispatchmethod` that was added in 3.8 as you noted in your comment (and your TODO). Can you make the switch to using that implementation for dispatch?

tests/program_analysis/CAST2GrFN/test_cast_json.py
dpdicken: I know that this is an option, however the tests are parsing the CAST from a json string and also comparing the json string from the file to the json string generated by the "to_json_str" method. AFAIK json.load returns an object if I am not mistaken. Perhaps they should be comparing objects instead and this change would make sense.

tests/program_analysis/CAST2GrFN/test_cast_json.py
dpdicken: sometimes a five minute google search is worth it haha

tests/program_analysis/CAST2GrFN/test_cast_json.py
pauldhein: > Perhaps they should be comparing objects instead and this change would make sense.
I would definitely prefer for us to compare objects. That will likely provide us with richer diffs if/when anything does differ between the objects we are comparing. Is that not the case? Additionally, if we need more packages to make the object comparison easier then we can talk about including those in AutoMATES.

pauldhein (approved):

PR #187, opened 2021-02-22
Handling conjunctions

maxaalexeeva:
- new tests and rule fixes from @alicekwak's branch and mine (NB! new `Concept`s and conj test file )
- switching to universal enhanced dependencies (might switch back later, but there doesn't seem to be an issue)
- more different methods to replace keepLongest in some places:
   - one to get most complete vars, e.g., R(t) instead of separate R, t, R(t, and R(t); 
   - one to only apply keep longest in cases when var is the same (needed mainly to get rid of mentions with same args extracted with different similar rules, but also to get rid of the mentions that have concepts instead of vars)
 - handling certain kinds of definitions containing conjunctions (a lot of methods associated methods to both extract them properly and store discont char offsets in attachments)
 - allowing expansion on "acl:relcl" (adding it to valid outgoing), but disallowing expansion after expansion on that relation (by adjusting the `traverseIncomingLocal` method)---expansion after this dependency results in grabbing too large of a span
 - allowing max expansion hop up to 16 to capture some key definitions - 12 is too low in case of long lists of variables in parens in the same sentence (e.g., <part of def shared between all three vars><def1> (<var>),  <def2> (<var2>), <def3>(<var3>)) 
 - `text_input_format` arg replaces with serializer_name
 - added a new exporter (used when running the ExtractAndExport app to extract mentions---only this one uses the serializer that stores attachments)
 - when deserializing mentions, making sure the `paths` key is available


pauldhein (approved):
Wow. I didn't realize just how much would be involved in preparing the TR pipeline to handle conjunctions. Great work @maxaalexeeva! I found one simple change (see comment above in the PR thread) and then these changes are ready to merge as soon as all checks pass.

maxaalexeeva:
@pauldhein although I see there are failing tests on ci/dockercloud-stage and I can't remember what those are (I see they are not one of the `required` tests, but the Details link for that type of test is broken, so I am not sure what that's about)

PR #191, opened 2021-03-03
Lambda Expression Walker

pauldhein:
# Overview
- Implements a parser and walker for lambda expressions stored in the GrFN `LambdaNode`s
- Converts expressions to a tree using `ExpressionNode`s that include `children` lists with implied ordering of children for binary operators, comparators, calls, etc.
- Includes a `nodes2digraph` method that converts a collection nodes into a DiGraph that can then be output as a PDF for visual analysis
- Includes a script `/scripts/model_assembly/expression_walker.py` that takes a GrFN JSON as input, converts all lambda expressions found in the GrFN, and then outputs the `ExpressionNode` collection in a new JSON file


titomeister (commented):
This looks good to me! Most of the changes appear to be adding comment blocks, but the actual lines of code

dpdicken:
The actual changes look good! Are you switched to using the black python formatter? It seems like every PR we reformat some files (i.e. networks.py in this pr)

dpdicken (approved):

PR #194, opened 2021-03-14
refactoring align

maxaalexeeva:
- main changes:
  - separated param settings and units from the text var object; 
  - now param settings, param setting interval, and unit mention extractions are all stored as separate link elements (objects) in the align output json; 
  - they are all linked to the previously existing text_var/def mention objects (definition objects are now removed); 
  - linking from these three event types to text variables happens based on whether param settings and units are attached to identifiers or concepts in the extracted mentions (_**t** is measured in seconds_ vs _**time** is measured in seconds_)
  - added argument objects with spans in each extraction object (in addition to the full mention span)
- other:
  - rewrote interval param setting rules to incl info about excl and incl bounds
  - added methods to process units and param settings 
  - discontinuous char offset attachments now attach to argument it modifies instead of the full event or relation mention
  - updated the looksLikeAVariable filter to filter out more units that look like vars (eg m-1) and include vars like R(t)
  - SVO temporarily removed from text var objects - will be handled as a separate pr
  - simplified some of the code (e.g., the rehydrateLinkElement method); related: object strings are now json object strings instead of "::"-separated strings - makes code more readable and helps simplify the rehydrateLinkElement method)


pauldhein (approved):

maxaalexeeva:
Hi @pauldhein! Does your approval mean it's safe to merge or would you rather merge yourself? (eg if you want all the current pr's merged in a certain order)

pauldhein:
@maxaalexeeva Once review on a PR is done feel free to merge it to master, but if you know that the PR should wait for whatever reason before being merged to master then certainly hold off merging. I'm only going to be reviewing PRs from now on and I'll have the PR creators merge them since the creators have a better understanding of their branches.

maxaalexeeva:
@pauldhein, sounds good! Thanks!

PR #196, opened 2021-03-19
cast-to-pdf initial pull request

titomeister:
All the nodes are implemented, and have documentation. Some of the documentation is a bit redundant, sorry about that!


pauldhein (approved):
Great work @titomeister!

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
```diff
@@ -0,0 +1,449 @@
...
+                "style": "rounded",    
+                "fontname": "Menlo"
+            }
+        )
+        return A        
+
+    def to_pdf(self,filename):
+        """Creates a pdf of the generated agraph"""
+        A = self.to_agraph()
+        for node in A.iternodes():
```
dpdicken, line 101: could you move this to around line 80 so all of the style updates are in the same place?

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
```diff
@@ -0,0 +1,449 @@
...
+            node.attr["fontcolor"] = "black"
+            node.attr["style"] = "rounded"
+        A.edge_attr.update({"color": "#650021", "arrowsize": 0.5})
+        A.draw(filename+".pdf",prog="dot")
+
+    @singledispatchmethod
+    def visit(self, node: AstNode):
+        """Generic visitor for unimplemented/incorrect nodes"""
+        print("Not implemented: ",node)
```
dpdicken, line 110: could you make an exception class and raise it here instead?

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
```diff
@@ -0,0 +1,449 @@
...
+        if(node.values != []):
+            values = self.visit_list(node.values)
+        node_uid = uuid.uuid4()
+        self.G.add_node(node_uid, label="Dict")
+
+        # TODO: Make this prettier, somehow?
+        for n in keys + values:
+            self.G.add_edges(node_uid, n)        
+
+        #for (k,v) in list(zip(keys,values)):
```
dpdicken, line 211: can you remove this commented code

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
```diff
@@ -0,0 +1,449 @@
...
+        keys = []
+        values = []
+        if(node.keys != []):
+            keys = self.visit_list(node.keys)
+        if(node.values != []):
+            values = self.visit_list(node.values)
+        node_uid = uuid.uuid4()
+        self.G.add_node(node_uid, label="Dict")
+
+        # TODO: Make this prettier, somehow?
```
dpdicken, line 207: this loop looks fine to me :)

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
```diff
@@ -0,0 +1,449 @@
...
+    @visit.register
+    def _(self, node: ModelIf):
+        """Visits a ModelIf (If statement) node.
+        We visit the condition, and then the body and orelse 
+        attributes if we have any. They're all added to the Graph 
+        accordingly. The node's UID is returned."""
+        expr = self.visit(node.expr)
+        body = []
+        orelse = []
+        if(node.body != []):
```
dpdicken, line 322: Im curious, do you need these if checks or can you just do 

body = self.visit_list(node.body)

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
```diff
@@ -0,0 +1,449 @@
...
+        self.G.add_node(node_uid,label="If")
+        self.G.add_node(test_uid,label="Test")
+        self.G.add_edge(node_uid,test_uid)
+        self.G.add_edge(test_uid,expr)
+        
+        body_uid = uuid.uuid4()
+        orelse_uid = uuid.uuid4()
+
+        self.G.add_node(body_uid,label="If Body")
+        self.G.add_node(orelse_uid,label="Else Body")
```
dpdicken, line 338: This code doesn't necessarily represent if/elifs correctly. For instance:

```
if x == 1:
    ....
elif x == 2:
    ....
elif x == 3:
    ....
```

I think we discussed before this should be a single if node with "cond 1 test" and "cond 1 body", "cond 2 test" and "cond 2 body", etc. until all of the conditions are handled. Then, if there is an else node too, we can have this else body labelled node. The "orelse" field represents the else part of the expression if it is a list of many nodes OR the next condition in the else if if the list has one node that is an if statement in it.

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
```diff
@@ -0,0 +1,449 @@
...
+        self.G.add_node(node_uid, label="Return")
+        self.G.add_edge(node_uid, value)
+
+        return node_uid
+
+    @visit.register
+    def _(self, node: Module):
+        """Visits a Module node. This is the starting point for visiting
+        a CAST object. The return value isn't relevant here (I think)"""
+        return self.visit_list(node.body)
```
dpdicken, line 367: I think this PR might be missing your module/program level node additions. Just remember to add the module name onto the module node too!

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
```diff
@@ -0,0 +1,449 @@
...
+            self.G.add_edge(node_uid,n)
+
+        return node_uid
+
+    @visit.register
+    def _(self, node: String):
+        """Visits a String node. We add this node's string to the
+        graph and return the UID of this node."""
+        node_uid = uuid.uuid4()
+        self.G.add_node(node_uid,label="\""+node.string+"\"")
```
dpdicken, line 404: small suggestion, you could use pythons built in string formatting in which case you would have 
```
label=f"\"{node.string}\""
```
to achieve the same thing.

scripts/program_analysis/cast_to_agraph.py
```diff
@@ -0,0 +1,77 @@
...
+    v = Var(val=Name(name="exampleVar"), type="float")
+    n1 = Number(number=36.2)
+    n2 = Number(number=37.8)
+    op = BinaryOp(op="Add",left=n1,right=n2)
+    a = Assignment(left=v, right=op)
+    f = FunctionDef(name=Name(name="exampleFunction"), func_args=[], body=[a])
+    m = Module(name="ExampleModule", body=[f])
+    return CAST([m])
+
+def basic_function_def_and_assignment_if():
```
dpdicken, line 53: I think you can remove these three functions that create CAST objects because you have updated main to read in a json file. You could also remove all the CAST imports if you remove these functions.

dpdicken (changes requested):
Looks good, just have some small comments.

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
pauldhein: @titomeister to add some context to @dpdicken's request ... it would be useful to have an exception here that stops processing so that we know and can handle cases of unrecognized CAST nodes. Something like the following would be the Pythonic implementation:
```python
def visit(self, node: AstNode):
    """Generic visitor for unimplemented/incorrect nodes"""
    raise TypeError(f"Unrecognized node type: {type(node)}")  
```

We want to show the user the `type` of `node` here because that is the piece of information used by `singledispatch()` to correctly route the call to `visit()`. Also, f-string (`f""`) syntax is very useful for creating strings composed with variables!

Additionally, Daniel's suggestion would help make this even more clear by defining a special error class that would describe the kind of error we would expect here (something like `CASTTypeError`) which could be raised in the place of the `TypeError` in my example.

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
pauldhein: @titomeister the function you are looking for is `add_edges_from` (documented [here](https://networkx.org/documentation/stable//reference/classes/generated/networkx.Graph.add_edges_from.html)). You can combine that with a [generator expression](https://docs.python.org/3/howto/functional.html#generator-expressions-and-list-comprehensions) to form the following Pythonic statement:
```python
self.G.add_edges_from((node_uid, n) for n in keys + values)
```

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
```diff
@@ -0,0 +1,449 @@
...
+
+    @visit.register
+    def _(self, node: Call):
+        """Visits Call (function call) nodes. We check to see
+        if we have arguments to the node and act accordingly. 
+        Appending all the arguments of the function to this node,
+        if we have any. The node's UID is returned."""
+        func = self.visit(node.func)
+        args = []
+        if(node.arguments != []):
```
pauldhein, line 166: @titomeister this is not the safest way to do this check. If a non-empty non-list type were ever assigned to `node.arguments` this check would still fail. A better version of the check would be to see if `node.arguments` is empty by doing the following:
```python
if len(node.arguments) > 0:
    ...
```

Please convert this check to use the version above and all other similar checks found in this file. Also, parens are not required around the conditional portion of an if-statement in Python and they are considered _non-Pythonic_ so I would appreciate it if they were removed 🙂.

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
pauldhein: I would also prefer to use the method @dpdicken described above.

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
pauldhein: Piggy-backing off this ... @titomeister Can we include the name of the Module in the module node? Right now I believe the output has the module node named `Module: module`. Am I mistaken about that?

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
pauldhein: Feel free to do this as well (`black` will reformat the code like this anyways):
```python
label=f'"{node.string}"'
```

pauldhein (changes requested):
Most of my change requests mirror the ones from @dpdicken. Definitely take a look at whether the _is utterable empty_ checks are necessary ... my guess is that if we recurse to a node with an empty literal nothing will happen so the check(s) _may_ be unnecessary. If they are necessary though, please change them to the form I suggested in my comment.

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
titomeister: Thanks Daniel :--) 
Thank you Paul! I'm going to add this later because I want to experiment with it a little bit before committing it to the code. I made myself a note about it so I won't forget. I'll go ahead and remove the TODO comment from here for now though.

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
titomeister: I put them in there because I recall having issues in other places when the checks weren't there, so I left them in just to be cautious. I can investigate more later and see if I truly need them, but I'd like to leave them there for now.

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
titomeister: Done! I changed them all to reflect this Pythonic version. Thank you for the suggestion. I'm not sure where I picked up the habit of using parentheses in my if statements: school or from programming in C...

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
titomeister: You're right, it is missing. I must've not pushed it, but I wrote it in after our discussions. Sorry about that!
Paul: Right, the PDFs have names for both the Program and Module nodes. Right now they're both set to be the same name, the name of whatever is in the 'name' field of the Module node.

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
titomeister: Will do! I'm still used to the more complicated (and uglier) method of manually concatenating strings. Thanks!

scripts/program_analysis/cast_to_agraph.py
titomeister: Done!

automates/program_analysis/CAST2GrFN/visitors/cast_to_agraph_visitor.py
titomeister: Right, I understand what you mean when it comes to how the visualization should look. I keep getting stuck on how to work the with the If CAST as it is to produce this. I'm going to leave this as it for the moment, and look into this more later.

dpdicken: ok, can you add a comment with a TODO about this, because it can only handle IF/IF else statements correctly as written. It would be good to just track that here.

dpdicken (changes requested):

pauldhein (approved):
It looks like all of our requests have been addressed. @dpdicken can you please take a look again and mark off all conversations which have been resolved and then either add your approval or request more changes? @titomeister great work!

dpdicken (approved):

PR #201, opened 2021-03-20
Global var

maxaalexeeva:
- groups text var/def objects by variable into global variable objects
- alignment to global variable objects happens either through the variable text or through the text of all definitions associated with the global variable

### DIFF
--- a/automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
+++ b/automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
@@ -20,10 +20,8 @@ import ujson.{Obj, Value}
 import org.clulab.grounding.{SVOGrounder, sparqlResult}
 import org.clulab.odin.serialization.json.JSONSerializer
 import java.util.UUID.randomUUID
-
 import org.clulab.aske.automates.attachments.{AutomatesAttachment, MentionLocationAttachment}
 import org.clulab.utils.AlignmentJsonUtils.GlobalVariable
-
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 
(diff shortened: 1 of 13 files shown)


maxaalexeeva:
@pauldhein, shoot, I forgot to tag you for review :'( Sorry!

automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
```diff
@@ -866,8 +902,8 @@ object ExtractAndAlign {
     val hypotheses = new ArrayBuffer[ujson.Obj]()
     val linkElKeys = linkElements.keys.toSeq
     // Comment -> Text Var
-    if (linkElKeys.contains(COMMENT) && linkElKeys.contains(TEXT_VAR)) {
-      hypotheses.appendAll(mkLinkHypothesis(linkElements(COMMENT), linkElements(TEXT_VAR), COMMENT_TO_TEXT, alignments(COMMENT_TO_TEXT), debug))
+    if (linkElKeys.contains(COMMENT) && linkElKeys.contains(GLOBAL_VAR)) {
```
pauldhein, line 905: @maxaalexeeva this function (`getLinkHypotheses`) contains many checks of the following form:
```scala
linkElKeys.contains(<some_key>)
```
The issue here is that we are checking if a key to a map is contained in a list of keys, instead of checking if the key exists in the map. Both of these checks will return identical results, but checking if the key exists in the map will take _O(1)_ time; whereas, checking if the key is in a list of keys will take _O(n)_ time. Luckily the fix is very easy. All of the checks in this function can be changed to the following:
```scala
linkElements.contains(<some_key>)
```
... that check will ensure that the key is in the map and will run in _O(1)_ time as documented [here](https://www.geeksforgeeks.org/scala-map-contains-method-with-example/). Can you please convert all `.contains()` checks in this function using the pattern above?

Additionally, the following check is run **many** times in this function:
```scala
linkElKeys.contains(GLOBAL_VAR)
```
Ideally, we only want to run this check once. Can you perform the `.contains()` check for _Global Var_ in an outer if-statement and then perform all subsequent `.contains()` checks for elements that link to _Global Var_ under that if-statement? That will greatly reduce the number of redundant checks happening in this function.

automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
```diff
@@ -878,46 +914,46 @@ object ExtractAndAlign {
...
 
     // TextVar to ParamSetting (through var)
-    if (linkElKeys.contains(TEXT_VAR) && linkElKeys.contains(PARAM_SETTING_THRU_VAR)) {
-      hypotheses.appendAll(mkLinkHypothesis(linkElements(TEXT_VAR), linkElements(PARAM_SETTING_THRU_VAR), TEXT_VAR_TO_PARAM_SETTING, alignments(TEXT_VAR_TO_PARAM_SETTING), debug))
+    if (linkElKeys.contains(GLOBAL_VAR) && linkElKeys.contains(PARAM_SETTING_THRU_VAR)) {
+      hypotheses.appendAll(mkLinkHypothesis(linkElements(GLOBAL_VAR), linkElements(PARAM_SETTING_THRU_VAR), TEXT_VAR_TO_PARAM_SETTING, alignments(TEXT_VAR_TO_PARAM_SETTING), debug))
     }
 
     // TextVar to ParamSetting (through concept)
     if (linkElKeys.contains(TEXT_VAR) && linkElKeys.contains(PARAM_SETTING_THRU_CONCEPT)) {
```
pauldhein, line 944: @maxaalexeeva I have a feeling the check here (line 944) was supposed to be `linkElKeys.contains(GLOBAL_VAR)`. Does this need to be fixed?

pauldhein (changes requested):
@maxaalexeeva thank you for the effort you have put into transitioning the TR pipeline towards using global text variables! I have some issues that need to be addressed (but the most important ones stem from old code that *needs* to be updated). Once those changes are made this branch is ready to merge!

pauldhein (approved):
This all looks great now. Thank you @maxaalexeeva for making those changes!

pauldhein:
@maxaalexeeva this branch is ready for you to hit the merge button as soon as you are ready!

maxaalexeeva:
Thanks, Paul! Will merge later tonight.

PR #206, opened 2021-03-31
Functions + conjunctions + TR debug

maxaalexeeva:
Merged a couple of branches @alicekwak and I have been working on. 

Major changes:
- initial function rules and tests 
- can now test conjunction extractions with discontinuous text (added a method to get the text of mention arguments taking into account the discontinuous span attachment), so now there are passing conjunctions tests
- handling another type of conjunctions (no span overlap in conjoined definitions---calling these `ConjDefinitionType2` for now)
- some new filtering methods to improve quality of extractions
- major work on documenting issues in test definitions - some of the issues fixed

Minor updates:
- [update] a lot of various TR debugging
- updating some examples for rules
- adding the check from looksLikeAVariableWithGreek to the regular looksLikeAVariable
- rule and test updates
- deleting K from the list of units (we don't want to lose variables called K)

PR #208, opened 2021-04-06
Align experiment with cosmos json

maxaalexeeva:
- Adjusted alignment_experiment to take cosmos json as input 
- Adjusted cosmos_json_to_mentions endpoint to work with this change
- alignment_experiment main() extracts mentions if `{MODEL_NAME}-mentions.json` with mentions does not exist; otherwise, uses the mentions from the existing `{MODEL_NAME}-mentions.json` file


maxaalexeeva:
@pauldhein, this should merge easily after https://github.com/ml4ai/automates/pull/206 is in master, but looks like it interferes with your tests? (see comparison here https://github.com/ml4ai/automates/compare/Alice_Conjuction_Test...alignExperimentFromCosmos1) 

This should simplify running align. Let me know what you think!

automates/model_assembly/interfaces.py
```diff
@@ -16,12 +16,24 @@ def extract_mentions(self, doc_path: str, out_path: str) -> dict:
...
-        print(f"HTTP {res} for /pdf_to_mentions on {doc_path}")
+        if doc_path.endswith(".pdf"):
+            res = requests.post(
+                f"{self.webservice}/pdf_to_mentions",
+                headers={"Content-type": "application/json"},
+                json={"pdf": doc_path, "outfile": out_path},
+            )
+            print(f"HTTP {res} for /pdf_to_mentions on {doc_path}")
+
+        elif doc_path.endswith(".json"):
```
pauldhein, line 27: @maxaalexeeva let's make this a little more specific.

I am creating all of the JSON files that include COSMOS processing output with the file-ending : `--COSMOS-data.json` (see examples of this in `ASKE-AutoMATES/Data/TR/COSMOS-pdf-output/`). Let's use that in the check here by doing the following:
```python
if ...:
  ...
elif doc_path.endswith("--COSMOS-data.json"):
    """< do COSMOS data processing here >"""
elif doc_path.endswith(".json"):
 """< do processing on non-COSMOS JSON here (such as the JSON saved results from parsing the ASCE paper with ScienceParse) >"""
else:
  raise ValueError(...)
```

This will ensure that we can still handle JSON output from ScienceParse while allowing us to distinguish that output from COSMOS processing output.

pauldhein (changes requested):
Just one change request dealing with how we determine COSMOS vs non-COSMOS JSON input to mention extraction.

automates/model_assembly/interfaces.py
maxaalexeeva: @pauldhein, great call!

scripts/model_assembly/alignment_experiment.py
```diff
@@ -7,12 +7,17 @@
 
 def main(args):
     CUR_DIR = os.getcwd()
-    MODEL_NAME = args.grfn_file.replace("--GrFN.json", "")
+    MODEL_NAME = os.path.basename(args.grfn_file).replace("--GrFN.json", "")
```
maxaalexeeva, line 10: @pauldhein, do you have any objections to getting the model name this way? otherwise, I can't pass the grfn file that is outside of the current directory.

maxaalexeeva:
@pauldhein, I addressed your feedback and also slightly modified an endpoint for extracting mentions directly from science parse-produced json.

scripts/model_assembly/alignment_experiment.py
pauldhein: I'm okay with this.

pauldhein (approved):

PR #209, opened 2021-04-11
Adding PA pipeline metadata to GrFN

pauldhein:
## Changes
- Introduces `automates/model_assembly/metadata.py` which holds a class hierarchy of:
  - typed metadata objects corresponding to the top-level metadata classes defined in [the GrFN Metadata Spec](https://ml4ai.github.io/automates-v2/grfn_metadata.html)
  - Helper metadata classes that help construct the typed metadata objects (but do not have provenance information)
  - Enum classes for critical attributes of various metadata classes
- Introduces `automates/model_assembly/air.py` that includes a new `AutoMATES_IR` class that has a `from_json` method to load AIR from a JSON file
- Updates `GroundedFunctionNetwork.from_AIR()` to use the new object created in `AutoMATES_IR`
- Updates tests as necessary to support the above items


automates/model_assembly/networks.py
```diff
@@ -819,10 +855,25 @@ def __init__(
...
+        del_indices = list()
+        for idx, var_node in enumerate(self.variables):
+            found_var = False
+            for edge in self.hyper_edges:
+                if var_node in edge.inputs or var_node in edge.outputs:
+                    found_var = True
+                    break
+            if not found_var:
+                self.remove_node(var_node)
+                del_indices.append(idx)
```
dpdicken, line 874: why don't you just delete the variable from self.variables here? (I think its because you cant modify the list while looping through it, but does enumerate create make a copy of this list so it doesn't matter?)

automates/model_assembly/networks.py
```diff
@@ -1006,7 +1060,8 @@ def translate_container(
                 func = add_lambda_node(
                     LambdaType.INTERFACE, interface_func_str
                 )
-                out_nodes = [add_variable_node(id) for id in con.arguments]
+                # out_nodes = [add_variable_node(id) for id in con.arguments]
```
dpdicken, line 1063: you can probably delete this

automates/model_assembly/networks.py
```diff
@@ -1057,16 +1112,17 @@ def _(
...
 
-            inputs = [live_variables[id] for id in stmt.inputs]
+            inputs = [variable_nodes[v_id] for v_id in stmt.inputs]
             (con_outputs, interface_func) = translate_container(
                 new_con, inputs, subgraph
             )
             Occs[stmt.call_id] += 1
-            out_nodes = [add_variable_node(var) for var in stmt.outputs]
+            out_nodes = [variable_nodes[v_id] for v_id in stmt.outputs]
+            # out_nodes = [add_variable_node(var) for var in stmt.outputs]
```
dpdicken, line 1125: can probably delete this

automates/model_assembly/networks.py
```diff
@@ -1057,16 +1112,17 @@ def _(
             live_variables: Dict[VariableIdentifier, VariableNode],
```
dpdicken, line 1112: does live variables still need to be passed around?

automates/model_assembly/networks.py
```diff
@@ -1079,9 +1135,10 @@ def _(
...
             subgraph: GrFNSubgraph,
         ) -> None:
-            inputs = [live_variables[id] for id in stmt.inputs]
-            out_nodes = [add_variable_node(var) for var in stmt.outputs]
-            func = add_lambda_node(stmt.type, stmt.func_str)
+            inputs = [variable_nodes[v_id] for v_id in stmt.inputs]
+            out_nodes = [variable_nodes[v_id] for v_id in stmt.outputs]
+            # out_nodes = [add_variable_node(var) for var in stmt.outputs]
```
dpdicken, line 1140: can probably delete this

automates/model_assembly/networks.py
```diff
@@ -1433,15 +1499,25 @@ def from_json(cls, json_path):
         H = [HyperEdge.from_dict(h, ALL_NODES) for h in data["hyper_edges"]]
 
         # TODO: fix this to actually gather typedefs
```
dpdicken, line 1501: I think this todo is resolved now?

dpdicken (changes requested):
A few minor comments but it looks good to me!

automates/program_analysis/CAST2GrFN/cast.py
```diff
@@ -124,13 +125,19 @@ def to_GrFN(self):
                     V[in_var] = VariableDefinition.from_identifier(in_var)
             C[new_container.identifier] = new_container
 
-        grfn = GroundedFunctionNetwork.from_AIR(
-            GenericIdentifier.from_str("@container::initial::@global::exampleFunction"),
+        # TODO: fix this to send types, objects and metadata
```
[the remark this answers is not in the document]
pauldhein: Yeah I'll remove types from this TODO. We definitely need to rethink documentation and represent it as metadata. Let's discuss that during our meetings next week. I'll add that to the TODO (as a sub-component of metadata).

automates/model_assembly/air.py
```diff
@@ -0,0 +1,67 @@
...
+from .structures import (
+    GenericDefinition,
+    VariableDefinition,
+    GenericContainer,
+    GenericIdentifier,
+    TypeDefinition,
+)
+
+
+class AutoMATES_IR:
```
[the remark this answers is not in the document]
pauldhein: Great! Let's keep those changes in mind for when you merge your code.

automates/model_assembly/metadata.py
```diff
@@ -0,0 +1,524 @@
...
+        return super().from_str(cls, data)
+
+
+@unique
+class CodeSpanType(AutoMATESBaseEnum):
+    VARIABLE_NAME = auto()
+    CODE_BLOCK = auto()
+    TYPE_DEF = auto()
+    TYPE_FIELD = auto()
+    FUNCTION_CALL = auto()
```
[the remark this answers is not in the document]
pauldhein: Yeah I do recall that now. We certainly can utilize an identifier type for both variables and type attributes. I wonder if we want something different from block/identifier to use for the function call line? Maybe that's still a block but a block with only one line. Thoughts?

automates/model_assembly/networks.py
pauldhein: Enumerate is only creating an iterator for the (mutable) object that uses the objects included `__next__()` method not a copy of the object and as you point out you don't want to modify a Python iterable during iteration.

automates/model_assembly/networks.py
pauldhein: Deleted!

automates/model_assembly/networks.py
pauldhein: Deleted!

automates/model_assembly/networks.py
pauldhein: I can't find this one so I guess someone else deleted it in some other branch that was then merged into this branch via master.

automates/model_assembly/networks.py
pauldhein: Yes it is! Thank you for catching that. Deleted!

automates/model_assembly/networks.py
pauldhein: We are still using `live_variables` to keep track of all the nodes included under a GrFNSubgraph:
```python
con_subgraph.nodes.extend(list(live_variables.values()))
```

We use those nodes in the `__call__` method of the GrFNSubgraph so I definitely think we need to keep the attr around and therefore we still need to use `live_variables`:
```python
if executed:
    all_nodes_visited.update(executed_visited_variables)
    all_nodes_visited.add(node_to_execute)
    node_execute_queue.extend(
        [
            succ
            for var in executed_visited_variables
            for succ in grfn.successors(var)
            if (
                succ in self.nodes
                and succ not in all_nodes_visited
            )
            or (
                var in self.nodes
                and succ.func_type == LambdaType.INTERFACE
            )
        ]
    )
else:
    node_execute_queue.extend(
        [
            lambda_pred
            for var_pred in grfn.predecessors(node_to_execute)
            for lambda_pred in grfn.predecessors(var_pred)
            if (
                lambda_pred in self.nodes
                and lambda_pred not in all_nodes_visited
            )
            or lambda_pred.func_type == LambdaType.INTERFACE
        ]
    )
    node_execute_queue.append(node_to_execute)
```

automates/model_assembly/metadata.py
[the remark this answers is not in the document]
dpdicken: right, I think our idea was any expression/multi line block of code can be labeled as a code block and any name can be labeled as an identifier. I.e. code block, function call, type def -> Code block while variable / function name -> identifier. For type fields, there would be a piece of meta data living on the variable found within the type that has an identifier referencing the name. I don't see why we would need a separate label for a function call vs any other expression.

dpdicken (commented):
Replied to some of your replies!

automates/model_assembly/metadata.py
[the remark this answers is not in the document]
pauldhein: @dpdicken it looks like we agree on this. Thank you for reminding me of the design decision we made as a group 🙂 

I have changed the `CodeSpanType` to only have the values `block` and `identifier`. All instances of code type generation have been updated based upon this change.

dpdicken (approved):
Looks good

dpdicken (approved):

PR #210, opened 2021-04-11
Adding literal overwrite option to GrFN execution

pauldhein:
## Changes
- Adds a `literals` argument to the `__call__` function in `GroundedFunctionNetwork` that allows users to specify values for literal variables that will override default values stored in GrFN
- Changes the specification method for inputs to a GrFN execution to be specified by full variable identifiers as opposed to variable short names
- Extends GrFNSubrgaph execution by adding an ability to continue execution even if no hyper-edges are found in the current subgraph by searching for child subgraphs to execution. Child subgraphs are ranked according to their contained functions maximum distance to an output node and the child subgraph with maximum rank is executed first.


dpdicken (approved):
I think the changes look good

PR #211, opened 2021-04-12
Adding 'ghost test' for cast_to_pdf

titomeister:
The 'ghost test' is a file I found in CAST2GrFN's tests called 'cast_all_nodes.json' which looks to have a good deal of the CAST nodes in the file. I think this is good enough to solve the coverage issues, but let me know and I could come up with something else. I wrote a test script that reads the JSON file, creates a pdf out of it, and then deletes the pdf. Then, the test passes by default.

### DIFF
--- a/automates/program_analysis/CAST2GrFN/cast.py
+++ b/automates/program_analysis/CAST2GrFN/cast.py
@@ -125,7 +125,9 @@ class CAST(object):
             C[new_container.identifier] = new_container
 
         grfn = GroundedFunctionNetwork.from_AIR(
-            GenericIdentifier.from_str("@container::initial::@global::exampleFunction"),
+            GenericIdentifier.from_str(
+                "@container::initial::@global::exampleFunction"
+            ),
             C,
             V,
             T,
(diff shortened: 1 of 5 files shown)


pauldhein (approved):
@titomeister I made some changes on this branch for you, but the way you implemented the _smoke test_ (I couldn't remember the correct name before) was entirely correct. Some changes I made:
- I formatted some code files tested here using _black_ so that they conform to the repo standard
- I fixed a few linter warnings that occurred in various files
- I added `from_json_file()` to the CAST object so we can load from JSON directly
- I changed they way you are constructing a CAST in your test to utilize the Pythonic classmethod construction paradigm
- I changed the way `to_pdf()` works in the `CASTToAGraphVisitor` class. It now expects a path to a PDF file instead of adding `.pdf` to the end of the input path (this is a preferred best practice).
- I updated your script for creating a CAST PDF such that it works with the updated `to_pdf()`
- I updated your test to have the assert check that the expected PDF file exists and was generated at the expected location specified in `pdf_filepath`.


Obviously this PR now has my approval, but do let me know in the discussion if you have any issues with any of the changes that I made!

PR #212, opened 2021-04-20
Major renaming

maxaalexeeva:
- Concept -> Phrase (what we have been calling _Concepts_ for rule writing are mainly phrases and we want to reserve the word _Concept_ to refer to variable descriptions that tell us what identifiers stand for (e.g., in _LAI stands for leaf area index_, where _leaf area index_ is the concept for the identifier _LAI_)
- Variable (label) -> Identifier (NB! variable **argument** stays the same (quoting @pauldhein: "Variables are abstract entities that are represented using identifiers, concepts, and definitions; therefore any attributable information about a variable (such as the units of a variable) can be associated with any of the variable representations (identifier, concept, or definition)" - so using the word _variable_ for this argument is okay because it covers both identifiers and concepts)
- definition -> description (both label and argument)
- same changes in all the relevant methods


pauldhein (approved):
Wow that was a lot of _very necessary_ name changes.

PR #214, opened 2021-04-29
Initial ID Algorithm Implementation

aswinchester:
## Additions
- Added two tests for id algorithm
- Added id_algorithm


automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,538 @@
...
+    :param x: nodes
+    :param y: nodes
+    :param z: nodes
+    :return: T/F if x is separated from y given z in g
+    """
+    if x == y:
+        return False
+    if len(x) == 0 or len(y) == 0:
+        return True
+    else:
```
pauldhein, line 367: @aswinchester the `if` statements above are acting as _guards_. That makes the `else` here feel a little weird. If I were you I would simply put the return call after the two _if-guards_ without adding an `else`.

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,538 @@
...
+                        stack[0:stack_size_old] = copy.deepcopy(stack_old)
+                        stack_names = [None] * stack_size
+                        stack_names[0:stack_size_old] = copy.deepcopy(stack_names_old)
+                    stack_add = stack_top + n_vis_pa + n_vis_ch
+                    stack[stack_top:stack_add] = [True] * n_vis_pa + [False] * n_vis_ch
+                    stack_names[stack_top:stack_add] = copy.deepcopy(visitable_parents) + copy.deepcopy(
+                        visitable_children)
+                    stack_top = stack_add
+            elif not el:
+                if not (el_name in z):
```
pauldhein, line 447: @aswinchester the operation `(el_name in z)` is a linear time operation. That's an expensive op for a conditional evaluation and you have now done it twice (in the cases where we reach the `elif`). Can you assign the result of that expression to a variable and then use that variable in both conditional evaluation?

automates/model_analysis/identification_algorithms.py
```diff
@@ -0,0 +1,251 @@
...
+    :param topo: topological ordering
+    :param tree: object to store line-by-line summary
+    :return: P(y | do(x)) if y is identifiable or failure.
+                Note that the output needs to be read by get_expression
+    """
+    to = None
+    frm = None
+    description = None
+    tree = gm.TreeNode()
+    if (len(p.var) == 0) and (not (p.product or p.fraction)):
```
pauldhein, line 88: @aswinchester ideally we want as little code duplication under conditionals as possible. Consider the assignments under this conditional. They have quite a lot in common and only a single difference. The following simplification will make the code a lot easier to read:

```python
if (len(p.var) == 0) and (not (p.product or p.fraction)):
  p = gm.Probability(var=v)

tree.call = gm.Call(y=y, x=x, p=p, g=g, line=0, v=v, id_check=False)
```
Notice that when we use this method of assigning the minimal number of constructs that need to be unique under a conditional we can actually eliminate the `else` branch entirely. Please make this change in your code.

automates/model_analysis/identification_algorithms.py
```diff
@@ -0,0 +1,251 @@
...
+            tree.call.id_check = True
+            return gm.ResultsInternal(p=p_prod, tree=tree)
+
+        # Line 7
+        tree.call.s = s_single
+        s_prime = None
+        for component in cc:
+            if set(s_single).issubset(set(component)):
+                s_prime = component
+                break
```
pauldhein, line 200: @aswinchester for safety I would add a check after this loop to see if `s_prime` was never assigned. That may save some heartache during future debugging if an example goes wrong. The following would work:
```python
if s_prime is None:
  raise RuntimeError("s_prime component not identified in cc")
```

automates/model_analysis/identification_algorithms.py
```diff
@@ -0,0 +1,251 @@
...
+            if p.product:
+                p_prod = gm.parse_joint(p, node, cond_set, v, topo)
+            else:
+                p_prod = deepcopy(p)
+                p_prod.var = [node]
+                p_prod.cond = cond_set
+            product_list.append(p_prod)
+        product_list.reverse()
+        x_new = gm.ts(set(x) & set(s_prime), topo)
+        if s_prime_length > 1:
```
pauldhein, line 220: @aswinchester similar to my comment about the other conditional with large duplicated commands above, let's simply this statement to be the following:
```python
if s_prime_length > 1:
  alex_name_me_pls = gm.Probability(product=True, children=product_list)
else:
  alex_name_me_pls = product_list[0]

nxt = compute_ID(
    y, x_new, alex_name_me_pls, g_s_prime, g_s_prime_obs, s_prime, topo, gm.TreeNode()
)
```

automates/model_analysis/identification_algorithms.py
```diff
@@ -0,0 +1,251 @@
...
+                             s_prime, topo, gm.TreeNode())
+        else:
+            nxt = compute_ID(y, x_new, product_list[0], g_s_prime, g_s_prime_obs, s_prime, topo, gm.TreeNode())
+        tree.children.append(nxt.tree)
+        tree.call.id_check = nxt.tree.call.id_check
+        return gm.ResultsInternal(p=nxt.p, tree=tree)
+
+
+def compute_IDC(y, x, z, p, g, g_obs, v, topo, tree):
+    if len(p.var) == 0:
```
pauldhein, line 231: @aswinchester see my above comment(s) about code duplication under conditionals and then update this instance and any other instances in this file.

pauldhein (changes requested):
@aswinchester this _clearly_ represents a lot of hard work and it is looking really great! I have some change requests that I would like you to address and then this work will be ready to merge.

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,538 @@
...
+        for edge in u:
+            edge_tuple = edge.tuple
+            if edge_tuple[1] > edge_tuple[0]:
+                edges_to_remove.append(edge.index)
+        g_copy.delete_edges(edges_to_remove)
+        e = g_copy.es.select(description="U")
+        e_len = len(e)
+        new_nodes = []
+        for i in range(e_len):
+            new_nodes.append("".join(["u_{", str(i + 1), "}"]))
```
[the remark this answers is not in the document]
aswinchester: This has been completed

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,538 @@
...
+        g_copy.add_vertices(new_nodes, attributes={"description": ["U"] * e_len})
+        edge_list = []
+        for i in range(e_len):
```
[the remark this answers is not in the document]
aswinchester: This has been completed

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,538 @@
...
+    Orders nodes by their topological order
+    :param nodes: Nodes to be ordered
+    :param topo_order: Order to arrange nodes
+    :return: Ordered nodes (indices)
+    """
+    node_set = set(nodes)
+    return [n for n in topo_order if n in node_set]
+
+
+def ancestors(node, g, topo):
```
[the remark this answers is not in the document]
aswinchester: This has been completed

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,538 @@
...
+    an_set = set()
+    for ans in an_list:
+        an_set = an_set | set(ans)
+    an_ind = list(an_set)
+    an_names = to_names(an_ind, g)
+    an = ts(an_names, topo)
+    return an
+
+
+def ancestors_unsort(node, g):
```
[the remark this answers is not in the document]
aswinchester: This has been completed

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,538 @@
...
+    :return: The set of edges in g that are not incoming to x or outgoing from z.
+    """
+    edges = set(g.es.select().indices)
+    to_x = set(g.es.select(_to_in=g.vs.select(name_in=x).indices).indices)
+    from_z = set(g.es.select(_from_in=g.vs.select(name_in=z).indices).indices)
+    selection = edges - to_x - from_z
+    return selection
+
+
+def get_expression(prob, start_sum=False, single_source=False, target_sym="^*("):
```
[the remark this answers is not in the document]
aswinchester: This has been completed

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,538 @@
...
+    :param g: graph
+    :param topo: topological ordering
+    :return: list of c-components (each c-component is a list of nodes)
+    """
+    a = g.get_adjacency()
+    n = a.shape[0]
+    v = g.vs["name"]
+    bidirected = []
+    for i in range(0, n):
+        for j in range(0, n):
```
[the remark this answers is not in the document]
aswinchester: This has been completed

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,538 @@
...
+    cc = []
+    cc_rank = []
+    for subgraph in subgraphs:
+        nodes = ts(subgraph.vs["name"], topo)
+        cc.append(nodes)
+        rank = 0
+        for node in nodes:
+            rank = rank + topo.index(node)
+        cc_rank.append(rank)
+    (cc_sorted, rank_sorted) = list(map(list, zip(*sorted(zip(cc, cc_rank), key=lambda ab: ab[1]))))
```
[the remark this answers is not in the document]
aswinchester: This has been completed

automates/model_analysis/graph_manipulation.py
aswinchester: This has been completed

automates/model_analysis/graph_manipulation.py
aswinchester: This has been completed

automates/model_analysis/identification_algorithms.py
aswinchester: This has been completed

automates/model_analysis/identification_algorithms.py
aswinchester: This has been completed

automates/model_analysis/identification_algorithms.py
aswinchester: This has been completed

automates/model_analysis/identification_algorithms.py
aswinchester: This has been completed

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,634 @@
...
+                visited_size = 2 * visited_size
+                visited = [False] * visited_size
+                visited[0:visited_size_old] = copy.deepcopy(visited_old)
+                visited_names = [None] * visited_size
+                visited_names[0:visited_size_old] = copy.deepcopy(visited_names_old)
+            visited[visited_top - 1] = el
+            visited_names[visited_top - 1] = el_name
+            el_name_in_z = el_name in z
+            if el and (not el_name_in_z):
+                visitable_parents = list((set(parents_unsort([el_name], g)) - set([el_name])) & set(an_xyz))
```
pauldhein, line 417: @aswinchester the code from line 417 -- 435 seems very similar to the code from line 438 -- 452. In fact, it seems the single difference rests on whether "parents" are included in the various calculations or not. Why don't you turn this code into a Python function scoped under the current function with an arg that controls whether "parents" are used or not and then cleverly design the singular function to perform the roles of both of the code blocks I outlined above. Here is an example of how to scope functions under functions in Python:
```python
def foo(a, b, parent_flag=False):
    def foo_helper():
        if not parent_flag:
            ...
       ...
```
Notice how functions scoped under a parent scope still have access to all of the variables in the parent scope. This should be helpful to you in designing your function for these two code blocks. Not only is this a good programming exercise but it will also increase the maintainability of your code. If you ever need to change these code blocks you will only have to do it in one place instead of two 🙂.

BTW, you can treat function names as variables in Python so it is perfectly acceptable for you to send the function names `children_unsort` and `parent_unsort` as values into a function. That may also be helpful for you when designing this function for your two blocks.

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,634 @@
...
+                            stack_size = 2 * stack_size
+                            stack = [False] * stack_size
+                            stack[0:stack_size_old] = copy.deepcopy(stack_old)
+                            stack_names[0:stack_size_old] = copy.deepcopy(stack_names_old)
+                        stack_add = stack_top + n_vis_ch
+                        stack[stack_top:stack_add] = [False] * n_vis_ch
+                        stack_names[stack_top:stack_add] = copy.deepcopy(visitable_children)
+                        stack_top = stack_add
+                if el_name in an_z:
+                    visitable_parents = list((set(parents_unsort([el_name], g)) - set([el_name]) & set(an_xyz)))
```
pauldhein, line 454: @aswinchester see my comment above and try to handle this block in that function as well.

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,634 @@
...
+    orig_name: str = None
+    val_assign: str = None
+    int_var: str = None
+
+
+class IDANotIdentifiable(Exception):
+    pass
+
+
+gamma = [CF("Y", "y", "X"), CF("X", "x_prime"), CF("Z", "z", "D"), CF("D", "d")]
```
pauldhein, line 624: @aswinchester please either move this code to a script that you can use for debugging/smoke-testing or move it to your testing file.

pauldhein (changes requested):
Hi @aswinchester I know you didn't request a review on this work (and thus I didn't give it a full review) but it has been a while since I provided feedback here and you did a great job improving the code from my last round of feedback so I figured I would add some more change requests. Please reach out to me on Slack if you need any clarifications about my comments!

automates/model_analysis/graph_manipulation.py
aswinchester: I wrote a helper function that is called three times in d-sep. This should address this and the following suggestion.

automates/model_analysis/graph_manipulation.py
aswinchester: (see above)

automates/model_analysis/graph_manipulation.py
aswinchester: I have moved this into a script :)

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,618 @@
...
+    :param g: graph
+    :return: Ancestors of nodes
+    """
+    an_list = g.neighborhood(node, order=g.vcount(), mode="in")
+    an_ind = list(set([a for ans in an_list for a in ans]))
+    an_names = to_names(an_ind, g)
+    return an_names
+
+
+def parents_unsort(node, g_obs):
```
pauldhein, line 102: @aswinchester it looks like you can combine this function and `children_unsort()` into the `ancestors()` function by adding two more arguments: `order` and `mode`. You could name that function:
```python
find_related_nodes_of(node, g, order, mode, topo=None)
```
I think this would be the best schema for this function since it will require you to define a `order` and `mode` for each function call and that will help document the call as to whether we are finding parents, children, or ancestors. Does that sound good to you?

pauldhein (changes requested):
Just one more function refactoring and then this PR is good to go!

automates/model_analysis/graph_manipulation.py
```diff
@@ -0,0 +1,618 @@
...
+    Finds all ancestors of a node and orders them
+    :param node: node (indicated by its index)
+    :param g: graph
+    :param topo: topological ordering
+    :return: Ancestors of node in topological ordering topo
+    """
+    an_list = g.neighborhood(node, order=g.vcount(), mode="in")
+    an_ind = list(set([a for ans in an_list for a in ans]))
+    an_names = to_names(an_ind, g)
+    an = ts(an_names, topo)
```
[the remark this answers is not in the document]
aswinchester: I have resolved this with the `find_related_nodes_of()` function

automates/model_analysis/graph_manipulation.py
aswinchester: I have created the function:
```
find_related_nodes_of(nodes, g, mode, order=1,  topo=None, exclude_orig=False):
```
and removed all of the old functions and updated their usages.

pauldhein (approved):
Thanks for making that final change @aswinchester. This branch is now ready to merge into master!

PR #215, opened 2021-05-12
filter function fix

maxaalexeeva:
adjustments to the filterFunction method (checking if there is an identifier before filtering on identifier label); modified taxonomy to include measurements as a phrase subtype

### DIFF
--- a/automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/taxonomy.yml
+++ b/automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/taxonomy.yml
@@ -3,11 +3,11 @@
       - Identifier:
         - VariableGazetteer
       - Model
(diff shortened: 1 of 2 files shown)


maxaalexeeva:
@alicekwak, the checks are failing because of the three failing function tests, but it should be save to merge.

alicekwak:
> @alicekwak, the checks are failing because of the three failing function tests, but it should be save to merge.

Hey @maxaalexeeva, these changes all look great! I just checked all the function tests and fixed the labels (passing or failing) if they are wrong. I also pushed these changes, so I hope the issue has been addressed. If it hasn't, please let me know!

maxaalexeeva:
@alicekwak, cool! You can merge this pr and then pull your Alice_functions branch to have these changes locally.

PR #216, opened 2021-05-16
Improvements on function extraction

alicekwak:
**Major changes**

- Added new function rules and tests (including more "generalized" rules which embrace several rules sharing similar syntactic structure)
- Edited "filterFunction" to filter out inputs/outputs with overlaps within the same mention.
- Edited "keepLongest" to allow function mentions to have overlapping inputs/outputs when their triggers are different.
- Edited expansionHandler so that it can be applied distinctively depending on the arguments' type (i.e. description vs. function)
- Changed the hierarchy of taxonomy so that Measurements(Value and Unit) are also captured as Phrases

**Minor change**

- Added a few definition tests


automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
```diff
@@ -96,31 +140,189 @@ class TestFunctions extends ExtractionTest {
...
-}
+  // Tests from 2003-DSSAT
 
+  val t1f = "the soil water content of each layer is updated by adding or subtracting daily flows of water to or from the layer due to each process."
+  failingTest should s"find functions from t1f: ${t1f}" taggedAs(Somebody) in {
+    val desired = Seq(
+      "the soil water content of each layer" -> Seq("daily flows of water", "layer"),
+      "the soil water content of each layer" -> Seq("daily flows of water", "layer")
+    )
+    val mentions = extractMentions(t1f)
```
maxaalexeeva, line 158: Hm... I wonder if we should think of adding the type of relation involved in the function as an attachment or something... (Could decide based on the trigger). Like here, there are two identical-looking (from the point of view of test) extractions. If there were an attachment with the type of relation involved, they would be different and it would make more sense (that would involved some modifications to how we test these and also some additional work on attachments, but that shouldn't be too bad).

With the way we test events like this, there are two checks:
- number of extractions is correct
- every 'desired' mention exists in the extracted mentions

The first one is good, but the problem with the second on is that one of the two extracted mentions can be completely different from desired and the test will still pass (because both of the desired are identical, so there has to be only one mention with the desired arguments for the test to be like "yep, mentions does contain an event with these arguments!"). I wonder if right here we can throw another check right here (for this specific test) that would say that text of the first one of the `mentions` is equal to the second one.

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
```diff
@@ -96,31 +140,189 @@ class TestFunctions extends ExtractionTest {
...
+    val mentions = extractMentions(t8g) // fixme: when the right side of an equal sign is only a numerical value, the equation should not be captured as a function, but as a parameter setting.
+    testFunctionEvent(mentions, desired)
+  }
+
+  // Tests from paper: 1982-Comparison-of-PT and Penman Methods-petpt and petpen_PM_and_PT
+
+  val t1h = "In both cases Et was derived indirectly from calculated values of open water evaporation(Eo), using the equation: Et = f Eo where f was an empirical factor."
+  failingTest should s"find functions from t1h: ${t1h}" taggedAs(Somebody) in {
+    val desired = Seq(
+      "Et" -> Seq("calculated values of open water evaporation(Eo)"),
```
maxaalexeeva, line 308: @alicekwak I wonder if we want 'calculated values' here?

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
```diff
@@ -96,31 +140,189 @@ class TestFunctions extends ExtractionTest {
...
+    val desired = Seq(
+      "Wind speed" -> Seq("canopy density"), // fixme: same rule does not extract more than one set of input/output. -> fixed
+      "Rno" -> Seq("canopy density"),
+      "vapor pressure deficit" -> Seq("canopy density")
+    )
+    val mentions = extractMentions(t5g)
+    testFunctionEvent(mentions, desired)
+  }
+
+  val t6g = "The evaporation rate in this special case Esx is first approximated as equal to 0.8P."
```
maxaalexeeva, line 280: Not sure if this is a function. Let's discuss in the meeting.

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
```diff
@@ -96,31 +140,189 @@ class TestFunctions extends ExtractionTest {
...
+  passingTest should s"find functions from t2g: ${t2g}" taggedAs(Somebody) in {
+    val desired = Seq(
+      "Es" -> Seq("supply of energy to the surface"), // fixme: one additional, unwanted concept is captured as an input due to bad parsing.
+      "water movement to the evaporating sites near the surface" -> Seq("hydraulic properties of the soil")
+    )
+    val mentions = extractMentions(t2g)
+    testFunctionEvent(mentions, desired)
+  }
+
+  val t3g = "The model was used to obtain the total evaporation rate E = Es + Ep of a developing grain sorghum (Sorghum bicolor L.) canopy in central Texas."
```
maxaalexeeva, line 250: Do we want to extract the descriptions in functions? I can see it going both ways (e.g., to disambiguate E from other Es vs. "well, why extract descriptions if we extract them elsewhere?")

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
```diff
@@ -96,31 +140,189 @@ class TestFunctions extends ExtractionTest {
...
+    val mentions = extractMentions(t5f)
+    testFunctionEvent(mentions, desired)
+  }
 
+  val t6f = "Hourly canopy photosynthesis on a land area basis is computed from the sum of sunlit and shaded leaf contributions " +
+    "by multiplying sunlit and shaded leaf photosynthetic rates by their respective LAIs."
+  failingTest should s"find functions from t6f: ${t6f}" taggedAs(Somebody) in {
+    val desired = Seq(
+      "Hourly canopy photosynthesis on a land area basis" -> Seq("sunlit and shaded leaf contributions"),
+      "Hourly canopy photosynthesis on a land area basis" -> Seq("sunlit and shaded leaf photosynthetic rates", "their respective LAIs")
```
maxaalexeeva, line 206: Based on our discussion yesterday, should the desired extraction be a little more granular? like `sunlit leaf photosynthetic rates -> LAIs` and `sunlit leaf photosynthetic rates -> LAIs`?

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
```diff
@@ -34,16 +33,61 @@ class TestFunctions extends ExtractionTest {
     testFunctionEvent(mentions, desired)
   }
 
+  val t4a = "The actual vapor pressure can be measured or it can be calculated from various humidity data, such as measured dew point temperature, wet-bulb and dry-bulb temperature, or relative humidity and air temperature data."
+  failingTest should s"find functions from t4a: ${t4a}" taggedAs(Somebody) in {
+    val desired = Seq(
+      "The actual vapor pressure" -> Seq("various humidity data") // todo: how can we deal with pronouns (i.e. it)? + another issue is the "such as" part. (needs discourse parsing?)
```
maxaalexeeva, line 39: Based on this example, this heuristic from Dane's paper looks correct

> a generally trustworthy heuristic that theearliest named entity in the sentence is likely to be the an-tecedent of a pronoun if they match grammatically (Hobbs,1978).  (An Investigation of Coreference Phenomena in the Biomedical Domain, Bell et al, 2016)

This does not have to be in this PR, though.

maxaalexeeva (commented):
@alicekwak, I have left some comments. Some need to be addressed before merging the pr, but some other ones are just some things for us to discuss later (e.g., Dane's paper heuristic for coreference). Let me know if you'd like to meet and discuss any of this before you make changes! Thank you! This looks good---only small changes needed!

alicekwak:
> @alicekwak, I have left some comments. Some need to be addressed before merging the pr, but some other ones are just some things for us to discuss later (e.g., Dane's paper heuristic for coreference). Let me know if you'd like to meet and discuss any of this before you make changes! Thank you! This looks good---only small changes needed!

@maxaalexeeva, thank you for reviewing all these! I'll make sure to address the issues found before merging this PR. Also, let's discuss some of the issues in our next meeting.

automates/text_reading/src/main/resources/application.conf
```diff
@@ -93,7 +93,7 @@ apps {
 
 alignment {
   alignerType = "pairwisew2v"
-  w2vPath = "/local/path/to/vectors.txt" // a word embeddings file (e.g., GloVe) with this header: <len_of_vocab><space><embedding_size>, e.g., "6B 50" (no quotes); use of different sets of embeddings will require adjusting alignment similarity thresholds, e.g., commentTextAlignmentScoreThreshold
+  w2vPath = "/Users/alicekwak/repos/automates/automates/text_reading/vectors.txt" // a word embeddings file (e.g., GloVe) with this header: <len_of_vocab><space><embedding_size>, e.g., "6B 50" (no quotes); use of different sets of embeddings will require adjusting alignment similarity thresholds, e.g., commentTextAlignmentScoreThreshold
```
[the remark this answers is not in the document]
alicekwak: @maxaalexeeva, thanks for catching this! I've switched to the original path before creating this PR, but I guess I didn't push the commit where I made the change.

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/descriptions.yml
```diff
@@ -217,11 +217,11 @@ rules:
     pattern: |
       (@description: Phrase [!entity = "NUMBER" & !word = /\(|\)|,|=|\[|\]/]) "," ([word = "denoted"] [word = "as"])? @variable:Identifier (?! [lemma = "be"]) (?! [chunk = /I-NP|B-NP/])
       |
-      (@description: Phrase [!entity = "NUMBER" & !word = /\(|\)|,|=|\[|\]/])  [word = ","] @variable:Identifier [word = ","] (?! [lemma = "be"]) (?! [chunk = /I-NP|B-NP/])
+      (@definition: Phrase [!entity = "NUMBER" & !word = /\(|\)|,|=|\[|\]/])  [word = ","] @variable:Identifier [word = ","] (?! [lemma = "be"]|[word = "and"]) (?! [chunk = /I-NP|B-NP/])
```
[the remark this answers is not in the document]
alicekwak: @maxaalexeeva, I replaced all the "definition" with "description" in that rule. Thanks!

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/vars.yml
```diff
@@ -26,4 +26,16 @@ passive_mark_verbs: "marked|defined|indicated|characterized"
 
 transparent_nouns: "level|amount|quantit" # do we need these in general?
 
-# Todo: nmod_from "X will follow from Y"
\ No newline at end of file
+# Todo: nmod_from "X will follow from Y"
+
+# FUNCTION TRIGGERS
+
+functionTrigger1: "[lemma = influence]|[lemma = modify]|[lemma = govern]|[lemma = limit]|[lemma = control]|[lemma = dominate]"
```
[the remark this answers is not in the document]
alicekwak: @maxaalexeeva, I have some questions regarding this comment. Are the triggers automatically translated as lemmas even if I omit `lemma=` part? Also, if I want to keep some triggers as words rather than lemmas, I should keep them as `word=` forms. Is that correct?

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/functions.yml
```diff
@@ -2,66 +2,157 @@ vars: org/clulab/aske_automates/grammars/vars.yml
 
 rules:
 
-  # todo: the number of inputs is undecided. how to incorporate unspecified number of inputs into the rules? (for now, most of the rules are set to have two inputs)
-  # todo: a lot of inputs and outputs need to be expanded.
   # todo: some of the rules have very similar structure, though their trigger words differ. Would it be possible to capture them in a single rule?
```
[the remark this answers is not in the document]
alicekwak: It is deleted now, thanks!

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/functions.yml
```diff
@@ -70,117 +161,325 @@ rules:
...
     type: dependency
-    # action: ${action}
-    example: "Daily Rso is a function of the time of year and latitude."
+    action: ${action}
+    example: "Life cycle progress through any given phase depends on a physiological day accumulator as a function of temperature and day length, in many cases."
     pattern: |
-      trigger = [word = "function"] (?="of")
-      input:Phrase+ = nmod_of | nmod_of (nmod_of)? conj_and
-      output:Phrase = (nsubj | <nmod_as nsubjpass)
+      trigger = [lemma = "depend"]|[lemma = "dependent"] (?="on")
```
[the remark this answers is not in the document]
alicekwak: I'll go with the first version, just in case. Thanks :)

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/functions.yml
```diff
@@ -70,117 +161,325 @@ rules:
...
+    priority: ${priority}
+    type: dependency
+    action: ${action}
+    example: "This discrepancy is of little consequence, however, since the normal temperature, vapor pressure, and wind speed conditions existing during the growing season when the data used in developing (9) were taken (Ritchie, 1971) combine in (1) to be approximately equal to Rn0."
+    pattern: |
+      trigger = [lemma = "combine"]
+      input:Phrase+ = <acl conj_and?
+      output:Phrase = advcl_in nmod_to
+
+  # note: how to write a token rule for function extraction? (with multiple inputs)
```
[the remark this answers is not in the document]
alicekwak: Yes they are. I moved them to the unused_rules file as you suggested. Thanks!

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/vars.yml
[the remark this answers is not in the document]
maxaalexeeva: Good questions. I think the lemma part can go inside the rule; I forget the exact syntax but something like `trigger = [lemma=/${trigger_group1}/]`

You can do words instead of lemmas, yes. Can you give me an example for when you need that? Basically, everything you have inside slashes /.../ is a regex. You can have triggers in the vars file be parts of words and then in the rule, use ^ to indicate that the trigger should start with a certain letter sequence, eg (in vars.yml) `triggers_group1:"defin|influence"` can be used to capture both verbs and gerunds in the trigger definition with `trigger = [lemma=/^${trigger_group1}/]` 

We can discuss more in the meeting next week or stay for a few minutes after the TR meeting this Thursday.

automates/text_reading/src/main/scala/org/clulab/aske/automates/actions/ExpansionHandler.scala
```diff
@@ -22,32 +22,48 @@ class ExpansionHandler() extends LazyLogging {
...
 //    println(s"$proportion --> ${mention.sentenceObj.getSentenceText}")
     if (proportion > threshold) {
-      expandArgsWithSyntax(mention, state, validArgs)
+      expandArgsWithSyntax(mention, state, validArgs, expansionType)
     } else {
       expandArgsWithSurface(mention, state, validArgs)
     }
   }
 
+//  def expandFunctionArgs(mention: Mention, state: State, validArgs: List[String]): Seq[Mention] = {
```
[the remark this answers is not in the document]
alicekwak: Yes, they are deleted now.

automates/text_reading/src/main/scala/org/clulab/aske/automates/actions/ExpansionHandler.scala
```diff
@@ -471,5 +723,74 @@ object ExpansionHandler {
...
+  val INVALID_INCOMING_FUNCTION = Set[scala.util.matching.Regex](
+    //"^nmod_with$".r,
+    //    "^nmod_without$".r,
+    //    "^nmod_except$".r
+    //    "^nmod_despite$".r
+  )
+
+  // regexes describing valid outgoing dependencies
+  val VALID_OUTGOING_FUNCTION = Set[scala.util.matching.Regex](
+    //    "^amod$".r, "^advmod$".r,
```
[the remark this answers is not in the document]
alicekwak: They are gone now!

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
alicekwak: This is an excellent point! I've been thinking that function tests should check for the trigger (or type of relation) along with inputs and outputs, but wasn't sure how to fix them. Thanks for the great ideas! Let's discuss this in more detail during our next meeting.

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/vars.yml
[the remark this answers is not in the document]
alicekwak: Thanks for the clarification! I wanted some trigger groups (i.e. functionTrigger2 or 4) to be words than lemmas to exclude their verbal usages (if they have). I'll try to fine-tune the triggers by using regex. Thank you for the great advices :)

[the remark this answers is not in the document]
maxaalexeeva: @alicekwak, another related thing: you can add allowable parts of speech as well, e.g., if you want to only have base form of verbs, you can say something like `trigger=[lemma=/<whatever verbs you want>/ & tag = "VB"]` or if you want anything verb-y, you can do `tag = /^VB/` (tag name starts with "VB"). I think right now these part of speech (POS) tags are what we use in our system: https://www.ling.upenn.edu/courses/Fall_2003/ling001/penn_treebank_pos.html

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
alicekwak: We will keep this as it is now.

automates/text_reading/src/main/scala/org/clulab/aske/automates/OdinActions.scala
```diff
@@ -695,6 +705,53 @@ class OdinActions(val taxonomy: Taxonomy, expansionHandler: Option[ExpansionHand
...
+    val (functions, other) = mentions.partition(_.label == "Function")
+    for (f <- functions) {
+      val newInputs = new ArrayBuffer[Mention]()
+      val newOutputs = new ArrayBuffer[Mention]()
+      val newTrigger = new ArrayBuffer[Mention]()
+      for (argType <- f.arguments) {
+        val sameInterval = argType._2.groupBy(_.tokenInterval) // group by token intervals
+        for (s <- sameInterval) {
+          if (argType._1 == "input" ) {
+            if (s._2.toList.length == 1) {newInputs ++= s._2} // if there's only one input, return that
```
pauldhein, line 719: @alicekwak You have several calls to `s._2.toList.length` in this block and some of them could be successive given the ordering of the conditionals. Please change this code to use a single call to this function, set the value to a variable, and then use that variable in all of your conditional checks as needed.

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
pauldhein: @alicekwak and @maxaalexeeva was this conversation ever resolved?

pauldhein (commented):
@alicekwak This is great code! It looks like Masha has already signed off on the rule changes. I have only one change request for your code and then I will approve the merge. I left a comment in the code for the change that I'd like you to address. Sorry for delaying this review!

pauldhein (changes requested):
See above comment.

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestFunctions.scala
maxaalexeeva: @pauldhein yes, keeping it as is for the time being.

alicekwak:
> @alicekwak This is great code! It looks like Masha has already signed off on the rule changes. I have only one change request for your code and then I will approve the merge. I left a comment in the code for the change that I'd like you to address. Sorry for delaying this review!

@pauldhein No problem! Thanks so much for reviewing my PR. I'll let you know when I'm done changing the code as you suggested.

automates/text_reading/src/main/scala/org/clulab/aske/automates/OdinActions.scala
alicekwak: @pauldhein Thanks for the comment! I've addressed this issue (I believe) by setting `s._2.toList.length` as `numOfArgs` and used the variable when needed. If this is not the change you intended with the comment, please let me know. Thank you!

pauldhein (approved):

pauldhein:
Thanks for fixing that @alicekwak. This PR is now approved and ready to merge as soon as you are ready.

PR #219, opened 2021-06-14
adjusting extract and export to use cosmos loader and updating what prints to tsv

maxaalexeeva:
@alicekwak 

if you run the ExtractAndExport app (the one I modified; use intellij or you can do it with `sbt run` and then enter the number for ExtractAndExport app, should be 4?), it will output a tsv file with some useful information (doc name, sentence, mention label and type + all args with text for that mention). Types of output (in this case .tsv) are defined in exportAs in application.conf. We normally use the json, but this tsv is more usuful for manual analysis.

### DIFF
--- a/automates/text_reading/src/main/resources/application.conf
+++ b/automates/text_reading/src/main/resources/application.conf
@@ -74,7 +74,7 @@ apps {
 
   outputDirectory = "./"
 
-  exportAs = ["serialized", "json"] // no other formats yet specified, but can be added!
+  exportAs =  ["tsv"]//,"serialized", "json"] // no other formats yet specified, but can be added!
   loadMentions = true
   mentionsFile = "/local/path/to/PT-mentions.json"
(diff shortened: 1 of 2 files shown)


maxaalexeeva:
@alicekwak you just need to change this path to the directory where you have cosmos documents to extract from and on the next line the output directory: https://github.com/ml4ai/automates/blob/9cbf171d817541f762d22571d89611268dcb626a/automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndExport.scala#L36

alicekwak:
@maxaalexeeva Thank you, Masha! I merged this branch and tried running the ExtractAndExport app following your instruction. I think I've adjusted all the needed paths, but I keep encountering this error message: 
`ujson.Value$InvalidData: Expected ujson.Obj` 
Would you happen to know what this error message is about? Thanks always for your help!

Issue #203, opened 2021-03-26
TR improvement


maxaalexeeva:
- [ ] handle parameter settings expressed in natural language, e.g., `where b is a positive number`

maxaalexeeva:
@alicekwak @paul, if you notice some issues with TR extractions or any cases you think we should handle related to any currently extracted mentions (variables, definitions, parameter settings and intervals, or units), feel free to throw them in on this thread, esp if you think that is something you would like me to have a look at.

alicekwak:
**Issues for definition extraction**

- [ ] handle cases where multiple definitions are given for a variable. (i.e. t2b - kl is the water extraction rate, an empiric soil–root factor for the fraction of available water that can be supplied to the plant from each rooted soil layer, t4h - the limiting value R∞, called the total size of the epidemic which is the total number of people having the disease at the end of the epidemic.)
- [ ] handle cases where expansion on acl:relcl causes issues (i.e. t2c - "distance from the axial center (m), S is a sink or source term" is wrongly extracted as the definition of d21)
- [ ] handle cases where a variable's definition is separated by parentheses. (i.e. t2l - standardized reference crop evapotranspiration for short (ETos) or tall (ETrs) surfaces)

alicekwak:
**Issues for variable extraction**

- [ ] handle variables with special characters (i.e. R∞ in t4h).
- [ ] handle compound variables with complex structure (i.e. e°(Tmax) in t4l). needs to add cases like this to compound var rule and looksLikeAVar action.

alicekwak:
**Issues for function extraction**

- [ ] handle mentions with overlapping inputs (i.e. t6f: one variable input (LAI) & one concept input (respective LAI) overlapping with each other, but not having the exact same tokenInterval)
- [ ] need a separate expansionHandler for function rules (i.e. t3a: for function extraction, it should not expand on conj_and, but for definition extraction, it should.)

alicekwak:
**Issue for parameter setting**

- [ ] fractions with variables as denominators interfere with parameter setting
      ex1) If we take a = 0.47 and b = 0.103 , then 1 / a + 1 / _b = 11.83_. (italicized part captured as parameter setting)
      ex2) so the length of the latent period is distributed exponentially with mean equals to _e-bs ds = 1_ / b (italicized part captured as parameter setting)

alicekwak:
**Issues for parameter setting**

- [ ] parameter setting for phrases is not extracted. 
[...]

Issue #193, opened 2021-03-08
need a separate preprocessor for webapp


maxaalexeeva:
PassThroughPreprocessor still does some cleaning (e.g., doesn't allow sequences that are less than 60% letters excluding spaces), so the webapp fails on text like this: `then: dI (0) > 0`.

maxaalexeeva:
addressed in pr #225

PR #221, opened 2021-06-24
Tr improvement may2021

maxaalexeeva:
- three versions of preprocessing (fixes the web app failing on short, math heavy text spans bug from here: https://github.com/ml4ai/automates/issues/193): passThrough for web app, light for general text, and EdgeCase for science-parse processed table and content-page heavy pdfs

Handled some of the issues from https://github.com/ml4ai/automates/issues/203 and other issues:
- added rudimentary bigram alignment + a list of stop words 
- now handling compound identifiers (added rules + event variable arg is expanded to the longest overlapping identifier) + adjusted looksLikeAnIdentifier action to allow compound identifiers
- adjusted the description filtering such that we only lose the ones that overlap in the variable argument with longer descriptions
- adjusted some description tests to pass because the reasons for failing were not crucial and added some toy tests based on longer real-text tests to check for some crucial things
- some small rule adjustments


maxaalexeeva:
This is now part of a different pr (draft here: https://github.com/ml4ai/automates/pull/225)

PR #222, opened 2021-06-30
grfn2gromet TR metadata

dpdicken:
@pauldhein all the changes for TR metadata into grfn. A WIP but you can see the progress here.


pauldhein (changes requested):
@dpdicken you have a lot of example files in this PR all under `scripts/model_assembly/example-tr-data/...`. Can we move these to `tests/data/example-tr-data/...`? That would be a more canonical storage location for them and then we can update paths as necessary when we use these files in scripts, test code, or even when necessary in the main code.

Oh also, we established having no PDFs in the repo as a rule a while ago to prevent large PDFs from cluttering the workspace. We can keep a single PDF for testing purposes (the small SIR-simple model PDF would be ideal for this), but other than that we should remove the PDFs if possible.

PR #225, opened 2021-07-14
TR improvement + align unit tests

maxaalexeeva:
TR and align improvement:

- three versions of preprocessing (fixes the web app failing on short, math heavy text spans bug from here: https://github.com/ml4ai/automates/issues/193): passThrough for web app, light for general text, and EdgeCase for science-parse processed table and content-page heavy pdfs

Handled some of the issues from https://github.com/ml4ai/automates/issues/203 and other issues:
- added rudimentary bigram alignment + a list of stop words 
- now handling compound identifiers (added rules + event variable arg is expanded to the longest overlapping identifier) + adjusted looksLikeAnIdentifier action to allow compound identifiers
- adjusted the description filtering such that we only lose the ones that overlap in the variable argument with longer descriptions
- adjusted some description tests to pass because the reasons for failing were not crucial and added some toy tests based on longer real-text tests to check for some crucial things
- some new rules and rule adjustments
- using the correct align method for texts (the one with embeddings and not just edit distance)
- disabled edit distance for text comparison 
- disabled alignment on variable text (relevant arg is only description now)
- added initial thresholds for alignments
- displaying unicode characters as symbols in cosmos contents
- for now, just reading from body text, not other types of content in cosmos
- made extract and export a little more usable (can use either cosmos or science parse; have a better tsv output and a better output file name)

Align testing:

- added toy files (hybrid of chime and double epidemic models)
- added unit tests for the toy model
- a lot of methods for testing


pauldhein (changes requested):
This looks great @maxaalexeeva! I only have one comment and it may not be relevant, but please give it a look over!

maxaalexeeva:
@pauldhein, great point! Will do that!

maxaalexeeva:
@pauldhein, made the change.

pauldhein (approved):

pauldhein:
@maxaalexeeva ready for merge!

maxaalexeeva:
@pauldhein, yay! Thank you!

PR #229, opened 2021-07-25
small fix for processing attachments

maxaalexeeva:
found by is only for context attachments


alicekwak:
@maxaalexeeva Thank you for the fix! It is correct that "foundBy" should only be used for context attachments. I'd like to merge this branch into Alice_Function2, but it seems like some checks are failing. Do you have any idea why these checks are not successful?

maxaalexeeva:
Those are about code coverage. I don't think it matters right now because we're not merging into master. But if you'd like, you can just adjust the code in your branch like I did here instead of merging this and close this pr.

alicekwak:
Thanks Masha! I just adjusted the code in my branch just in case. I'll close this PR now.

PR #233, opened 2021-08-04
Wikidata summer2021

maxaalexeeva:


maxaalexeeva:
@pauldhein, there's a new library, which might be why everything fails? It would also be great if we could discuss my caching---there is a chance we do not need that library at all.

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/descriptions.yml
```diff
@@ -311,6 +311,25 @@ rules:
     pattern: |
       [word=/a|an|the/] (?<description> [tag=/JJ|VBN/] [tag=/^N/] ) @variable:Identifier (?![tag=/^N/])
 
+  - name: at_rate
```
alicekwak, line 314: If possible, could you add examples to the newly added rules? (at_rate, at_rate_with_verb, identifier_def_ideal_table) That way it would be clearer for me what these rules are for.

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestDescriptions.scala
```diff
@@ -269,7 +269,7 @@ class TestDescriptions extends ExtractionTest {
 
   val t1c = "A convenient soil hydraulic property that will be used in this study is the matric flux potential " +
     "Mh0 (m2 d-1)"
-    passingTest should s"find descriptions from t1c: ${t1c}" taggedAs(Somebody) in {
+    failingTest should s"find descriptions from t1c: ${t1c}" taggedAs(Somebody) in {
```
alicekwak, line 272: why did this test become failing?

automates/text_reading/src/main/resources/frequentWords.tsv
```diff
@@ -1,3 +1,4 @@
+US
```
alicekwak, line 1: Oh, this is a good addition :)

alicekwak (commented):
@maxaalexeeva, I just looked through the parts where you made changes related to TR rules or tests. They looked great, but it wasn't clear for me what your new definition rules are trying to capture. So if possible, could you add examples to them? Other than that, everything looked good to me. Great job!

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/descriptions.yml
maxaalexeeva: Added examples for the at_rate rules; getting rid of identifier_def_ideal_table---that was to specific for one of the papers. Good catch!

automates/text_reading/src/test/scala/org/clulab/aske/automates/text/TestDescriptions.scala
maxaalexeeva: That was because I disallowed an adjective in a definition rule and potential was found as an adjective. I changed that back for now, so this is passing again.

automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/descriptions.yml
alicekwak: Thanks for adding the examples! But I'm still not getting what we're aiming to extract as a definition here. What is extracted as the definition for _ls_ in this sentence?: Susceptible individuals acquire infection with strain 1 at a rate Ic.

alicekwak: Could you explain a little more about the context of that example sentence and the rule itself? Maybe in our meeting tomorrow?

maxaalexeeva: Basically, trying to extract as much information as possible to allow good alignment under the conditions of no real description, so ideally, I'd get "Susceptible individuals acquire infection with strain 1 at a rate" as description. I think because of expansion or something like that I grab the identifier itself as part of the description, but that's a whole other issue that needs to be fixed for some other cases too and was a necessary evil.

alicekwak: Oh, I see. That makes sense now. Thank you for the explanation! I was confused because I couldn't find any definition there. But I guess this rule is not for extracting real definitions but for extracting information needed for improving alignment.

.github/workflows/ci.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
           sudo apt-get -y --no-install-recommends install wget
           docker pull $TR_IMAGE
           wget http://vanga.sista.arizona.edu/automates_data/vectors.txt -O $GITHUB_WORKSPACE/automates/text_reading/src/test/resources/vectors.txt
+          cp $GITHUB_WORKSPACE/automates/text_reading/src/test/resources/vectors.txt $GITHUB_WORKSPACE/automates/text_reading/vectors.txt
```
pauldhein, line 26: @maxaalexeeva the `vectors.txt` file is only used for running `sbt test`. If you only need it to be located at `$GITHUB_WORKSPACE/automates/text_reading/vectors.txt` for the tests to pass then lets locate it there and not have it in two places. However, if it is needed in both locations then we can leave the file as it is currently.

automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
```diff
@@ -218,25 +235,42 @@ object ExtractAndAlign {
...
-  def getGlobalVars(descrMentions: Seq[Mention]): Seq[GlobalVariable] = {
+  def getGlobalVars(descrMentions: Seq[Mention], wikigroundings: Option[Map[String, Seq[sparqlWikiResult]]], groundToWiki: Boolean): Seq[GlobalVariable] = {
 
     val groupedVars = descrMentions.groupBy(_.arguments("variable").head.text.replace(".", ""))
     val allGlobalVars = new ArrayBuffer[GlobalVariable]()
     for (gr <- groupedVars) {
       val glVarID = randomUUID().toString()
       val identifier = gr._1//AlignmentBaseline.replaceWordWithGreek(gr._1, AlignmentBaseline.word2greekDict.toMap); for now, don't convert: text vars are already symbols and comments shouldnt be converted except for during alignment
+      val identifierComponents = if (identifier.contains("_")) {
+        if (!identifier.split("_").map(_.length).contains(1)) {
```
pauldhein, line 246: @maxaalexeeva splitting strings on special characters is an expensive command to run more times than necessary (especially since this will be done for many identifiers). Can you assign the result of `identifier.split("_")` to a value and then run multiple checks on that value?

automates/text_reading/webapp/app/controllers/HomeController.scala
```diff
@@ -199,11 +226,12 @@ class HomeController @Inject()(cc: ControllerComponents) extends AbstractControl
...
     val textsAndLocations = loader.loadFile(jsonPath)
-    val texts = textsAndLocations.map(_.split("::").head)
-    val locations = textsAndLocations.map(_.split("::").tail.mkString("::")) //location = pageNum::blockIdx
+    val texts = textsAndLocations.map(_.split("<::>").slice(0,2).mkString("<::>"))
+    val locations = textsAndLocations.map(_.split("<::>").takeRight(2).mkString("<::>")) //location = pageNum::blockIdx
 
+    println("started extracting")
     // extract mentions form each text block
-    val mentions = texts.map(t => ieSystem.extractFromText(t, keepText = true, filename = Some(jsonPath)))
+    val mentions = texts.map(t => ieSystem.extractFromText(t.split("<::>").head, keepText = true, Some(t.split("<::>").last)))
```
pauldhein, line 234: @maxaalexeeva this one isn't as big of a deal since testing won't involve tons of data processing, but it is still a best practice to make reusable variables for any expensive snippet of code that you would run more than once. Here you could make a variable for `t.split("<::>")` that would prevent a double computation.

(Side note: I'm coming from Python land where it's more necessary to do these types of optimizations. In Scala land your compiler may actually be doing this for you, but it is a best practice nonetheless!)

scripts/text_reading/beautify_cosmos_with_ftfy.py
```diff
@@ -0,0 +1,33 @@
+import argparse
+import ftfy
```
pauldhein, line 2: @maxaalexeeva a few things here...

(1) `ftfy` is a package that is available via PyPi. We have an import strategy for our Python code in the repo and that strategy involves separating imports based upon their origin. The schematic for our import strategy is shown below:

```python
import <package-from-python-StdLib>
import <package2-from-python-StdLib >
...
import <package-from-PyPi>
import <package2-from-PyPi>
...
import <package-defined-python-file-in-AutoMATES>
import <package2-defined-python-file-in-AutoMATES>
```
Can you please move the `ftfy` import so that it fits that style?

(2) you should add the import to our `setup.py` file with the other imports for the repository that way new members will have this package installed when they setup their VM. (While you are doing that, I think I forgot to add `fastparquet` to that list. If you could add that package as well that would be great!)

scripts/text_reading/beautify_cosmos_with_ftfy.py
```diff
@@ -0,0 +1,33 @@
...
+input_dir = args.cosmos_dir
+output_dir = args.output_dir
+
+
+for input_file in os.listdir(input_dir):
+    print(input_file)
+    if input_file.endswith("--COSMOS-data.json"):
+        print("using file")
+
+        f = open(os.path.join(input_dir, input_file))
```
pauldhein, line 19: @maxaalexeeva the way this code is written you are leaving a file pointer open till the Python script exits. If something bad happens during the Python script execution this could actually corrupt the file to where it is no longer usable. Please change this and the line below it to one of the two following options:
```python
loaded_json = None
with open(os.path.join(input_dir, input_file)) as infile:
    loaded_json = json.load(infile)
```
... or ...
```python
loaded_json = json.load(open(os.path.join(input_dir, input_file)))
```

Either one of those options prevents the file pointer from hanging around after loading from JSON. You could also do `f.close()` after line 20 but I always forget that option because I never use it. :-)

scripts/text_reading/beautify_cosmos_with_ftfy.py
```diff
@@ -0,0 +1,33 @@
...
+        for item in loaded_json:
+            # print("item: ", item)
+            item["content"] = ftfy.fix_text(item["content"])
+
+
+
+        # the json file where the output must be stored
+        out_file_path = os.path.join(output_dir, input_file.replace("--COSMOS-data.json", "--pretty--COSMOS-data.json"))
+        print("out file path: " + out_file_path)
+        out_file = open(out_file_path, "w")
```
pauldhein, line 32: @maxaalexeeva same story here, please ensure the file pointers are closed properly!

automates/text_reading/src/main/scala/org/clulab/grounding/WikiGrounder.scala
```diff
@@ -0,0 +1,168 @@
...
+        val res = WikidataGrounder.runSparqlQuery(term, WikidataGrounder.sparqlDir)
+        result.append(res)
+      }
+      fileCache(term) = result.head
+      val allSparqlWikiResults = new ArrayBuffer[sparqlWikiResult]()
+      if (result.head.nonEmpty) {
+        val lineResults = new ArrayBuffer[sparqlWikiResult]()
+        val resultLines = result.head.split("\n")
+        for (line <- resultLines) {
+          val splitLine = line.trim().split("\t")
```
pauldhein, line 72: @maxaalexeeva consider refactoring lines 72 -- 77 by using [iterable unpacking](https://riptutorial.com/scala/example/3070/tuple-extractors). That will likely make this code look a lot cleaner and more condensed. Plus it looks extra cool 😎

automates/text_reading/src/main/scala/org/clulab/grounding/WikiGrounder.scala
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  def editDistance(s1: String, s2: String): Double = {
+    val dist = LevenshteinDistance.getDefaultInstance().apply(s1, s2).toDouble
+    dist
+  }
+
+  def editDistanceNormalized(s1: String, s2: String): Double = {
+    val maxLength = math.max(s1.length, s2.length)
+    val levenshteinDist = LevenshteinDistance.getDefaultInstance().apply(s1, s2).toDouble
```
pauldhein, line 151: @maxaalexeeva replace this with a call to the function `editDistance()` defined above. It is common practice to reuse functions even for single lines of code. In the future, we may wish to edit the `LevenshteinDistance` call or extend it with additional steps. It's best to only have to make those changes in a single location instead of trying to keep two separate but exactly the same pieces of code in sync with each other!

pauldhein (changes requested):
Hi @maxaalexeeva!

I was finally able to finish my final review. Overall this is quite good, and I only have a few minor change requests. Since I won't be around to deal with the final approval, I suggest that you add @dpdicken as a reviewer who can sign off on this PR after you finish making the changes I requested here.

alicekwak (commented):
Thanks for addressing my comments, @maxaalexeeva! I didn't find any further issue from my side :)

automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndExport.scala
```diff
@@ -32,8 +33,9 @@ object ExtractAndExport extends App {
 
   val config = ConfigFactory.load()
 
-  val inputDir: String = ""
-  val outputDir: String = ""
+  val numOfWikiGroundings: Int = config[Int]("apps.numOfWikiGroundings")
+  val inputDir: String = "/Users/alexeeva/Desktop/automates-related/SuperMaaS-sept2021/cosmos-jsons-beautified"
```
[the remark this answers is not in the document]
maxaalexeeva: This has been long overdue. Been reluctant to disrupt my existing workflow (input/output dirs defined in the script instead of loaded from config) :D Fixing it now.

.github/workflows/ci.yaml
maxaalexeeva: I now also need vectors some place else in main, so things break down if they are not in both places (main and test) :(

automates/text_reading/src/main/scala/org/clulab/aske/automates/apps/ExtractAndAlign.scala
maxaalexeeva: Great catch! Thanks!

automates/text_reading/src/main/scala/org/clulab/grounding/WikiGrounder.scala
maxaalexeeva: I might keep it as is for now because this way it might be easier to keep track of which information is associated with which split line item and I anticipate the need to add more information from wikidata (all this assuming I understood the link correctly).

automates/text_reading/src/main/scala/org/clulab/grounding/WikiGrounder.scala
maxaalexeeva: good eye! thanks!

.github/workflows/ci.yaml
dpdicken: To follow up on this conversation, if you need to use this under tests and in the actual application, you should be able to have it just in the actual application and then it should be available to the tests too (At least that's how it works for java apps and I would assume its the same for Scala).

maxaalexeeva: Ah I will try that. I think I misunderstood what Paul said. Makes sense. I'll try to fix that today.

maxaalexeeva: @dpdicken fixed

maxaalexeeva:
Hi, @dpdicken! Just checking in on whether you think this branch is ready to merge. I think I addressed everything addressable (incl. not copying the vectors file unnecessarily).

dpdicken (approved):

PR #234, opened 2021-08-05
Add mocked gromet experiment execution endpoint

dpdicken:
Adds an endpoint to the automates app that mimics execution of a grommet experiment. It allows a user to give inputs and select outputs for CHIME-SIR and SIR-Simple grommets. 

When a particular gromet model is selected, the endpoint actually runs the python source code the gromet is generated from and returns the result.

### DIFF
--- a/automates/apps/automates/model_code/chime_sir.py
+++ b/automates/apps/automates/model_code/chime_sir.py
@@ -112,7 +112,8 @@ def execute(
 
 def drive(start, end, step, parameters):
 
-    (S, E, I, R) = execute(**parameters, n_days=end)
+    param_names_to_vals = {k.split("::")[-2]: v for k, v in parameters.items()}
+    (S, E, I, R) = execute(**param_names_to_vals, n_days=end)
 
     def process_time_step_results(result_arr):
         stepped_results = []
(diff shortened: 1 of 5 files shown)


automates/apps/automates/model_code/sir_simple.py
```diff
@@ -0,0 +1,75 @@
...
+            parameters["gamma"],
+            i,
+        )
+
+        step_results["S"].append(S)
+        step_results["I"].append(I)
+        step_results["R"].append(R)
+        step_results["dt"].append(i)
+
+    return step_results
```
pauldhein, line 19: @dpdicken I think the keys in the final returned results should be GroMEt port IDs. Using the generated gromet for Simple_SIR I think we should hard-code these values to be:
```
"SIR-simple::SIR-simple::sir::0::--::s::0"
"SIR-simple::SIR-simple::sir::0::--::i::0"
"SIR-simple::SIR-simple::sir::0::--::r::0"
```

automates/apps/automates/model_code/chime_sir.py
```diff
@@ -0,0 +1,129 @@
...
+    (S, E, I, R) = execute(**parameters, n_days=end)
+
+    def process_time_step_results(result_arr):
+        stepped_results = []
+        for i in range(start, end + 1, step):
+            stepped_results.append(result_arr[i])
+        return stepped_results
+
+    return {
+        "S": process_time_step_results(S),
```
pauldhein, line 124: @dpdicken I think the keys in the final returned results should be GroMEt port IDs. I think we should hard-code these values to be:
```
"CHIME_SIR::CHIME_SIR::main::0::--::S::0"
"CHIME_SIR::CHIME_SIR::main::0::--::I::0"
"CHIME_SIR::CHIME_SIR::main::0::--::R::0"
```

pauldhein (changes requested):
@dpdicken looks good. Only one change request with the names of the output fields. Can you address that?

pauldhein (approved):

PR #245, opened 2021-08-23
Fixes for GrFN metadata link table generation 

dpdicken:


pauldhein (approved):
These renamings seem fine as long as the tests pass.

PR #251, opened 2021-09-16
More flexible COSMOS Ingestion name matching

pauldhein:


alicekwak (approved):
@pauldhein, this looks great to me 👍 I'll approve it now!

PR #255, opened 2021-09-24
Various fixes for translating GCC ASTs into GrFN

dpdicken:
* Fixes FROM_SOURCE being set incorrectly
* Adds LITERAL_FUNCTION_ARG from source creation type
* Correctly propagates global variable code span information


titomeister (approved):
Looks good !!!! Nice job !!! :)

PR #261, opened 2021-10-08
Add Ghidra scripts

dpdicken:
Add two python Ghidra scripts into the code base that allow us to dump the P Code for a binary.


titomeister (approved):
Looks good!

PR #265, opened 2021-10-13
Breaks/Continues/Early Returns GCC -> CAST Pull Request

titomeister:
Pushing up the changes made up to this point for handling and testing breaks/continues/early returns.
What remains is to finish off the testing by adding in the correct CAST files to the test framework and to make sure that the break handling is correct.

### DIFF
--- a/automates/program_analysis/CAST2GrFN/visitors/cast_to_air_visitor.py
+++ b/automates/program_analysis/CAST2GrFN/visitors/cast_to_air_visitor.py
@@ -517,8 +517,10 @@ class CASTToAIRVisitor(CASTVisitor):
 
             called_func_name = parse_attr(node.func)
 
-        # Skip printf calls for now
-        if called_func_name == "printf" or called_func_name == "print":
+        # TODO Handle scanf calls as input??
+        # Skip printf and scanf calls for now
+        funcs_to_skip = {"printf", "print", "scanf"}
+        if called_func_name in funcs_to_skip:
             return []
         # TODO throw error if not found
         # TODO also what happens if func is defined after call
--- a/automates/program_analysis/GCC2GrFN/gcc_plugin_driver.py
+++ b/automates/program_analysis/GCC2GrFN/gcc_plugin_driver.py
@@ -3,16 +3,76 @@ import subprocess
 
 GCC_10_BIN_DIRECTORY = "/usr/local/gcc-10.1.0/bin/"
 
-def run_gcc_plugin(language, input_files, plugin_location="./"):
+
+def run_gcc(
+    compiler_name,
+    full_plugin_path,
+    input_files,
+    compile_binary,
+    no_optimize,
+):
+    command = [
+        f"{GCC_10_BIN_DIRECTORY}/{compiler_name}",
+        f"-fplugin={full_plugin_path}",
+    ]
+
+    if no_optimize:
+        command.append("-O0")
+
+    if not compile_binary and len(input_files) == 1:
+        command.append("-c")
+
+    command.extend(input_files)
+
+    if not compile_binary:
+        command.extend(["-o", "/dev/null"])
+
+    # Runs gcc with the given source file. This should create the file ast.json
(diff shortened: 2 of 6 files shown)


automates/program_analysis/GCC2GrFN/gcc_ast_to_cast.py
```diff
@@ -514,12 +520,49 @@ def parse_conditional_statement(self, stmt, statements):
             # this second condition. Otherwise, we only have one condition and
             # the "false block" is the normal block of code following the if,
             # so do not evaluate it here.
+            # In the case that a block has a break in it, the first check won't work
+            # because the break changes the exit target of the block it's in, so we do 
+            # additional checks to see where the break takes us 
             if true_exit_target == false_exit_target:
                 false_res = self.parse_basic_block(false_block)
+            elif true_exit_target != false_exit_target and true_exit_target > true_block["edges"][0]["source"]:
```
dpdicken, line 528: Technically we already know `true_exit_target != false_exit_target` is true in this statement because `true_exit_target == false_exit_target` failed!

automates/program_analysis/GCC2GrFN/gcc_ast_to_cast.py
```diff
@@ -514,12 +520,49 @@ def parse_conditional_statement(self, stmt, statements):
...
 
             self.current_basic_block = temp
             return [ModelIf(expr=condition_expr, body=true_res, orelse=false_res)]
 
+    def parse_predict_stmt(self, stmt):
+        if stmt["name"] == "continue":
+            return [ModelContinue()]
+        if "early return" in stmt["name"]:
+            curr_b = self.current_basic_block
+            ret_val = curr_b["statements"][0]["operands"][0]
```
dpdicken, line 540: I don't think this always works / is the correct return value. Consider:

```
int main () {
    int x = 1;

    if (x == 2) {
        x += 1;
        return x * x + 2;
    }

    return x;
}
```

Theres two issues. One, in this scenario, `curr_b["statements"][0]` value will be the statement `x += 1` which doesn't have anything to do with the return statement. Second, you cannot just select `operands[0]` because there could be multiple operands to the return expression. For example, in the code snippet above, there is x and the constant value 2 as operands.

So, I think you need to identify the correct statement to parse (I think `curr_b["statements"][-1]` ?) and then parse all the operands and figure out which one is the actual related return value.

automates/program_analysis/GCC2GrFN/gcc_ast_to_cast.py
```diff
@@ -514,12 +520,49 @@ def parse_conditional_statement(self, stmt, statements):
...
+
+    def parse_goto_stmt(self, stmt):
+        
+        # Attempt to find which block this goto statement is from
+        if len(self.loop_exits) > 0:
+            loop_indices = list(self.loop_exits.keys())
+            curr = self.loop_exits[loop_indices[0]]
+            target = stmt["target"]
+
+            for loop_index in self.loop_exits:
```
dpdicken, line 554: Instead of saying 
```
            for loop_index in self.loop_exits:
                 loop_block = self.loop_exits[loop_index]
                 ...
```

You can say
```            
for loop_index, loop_block in self.loop_exits.items():
    ...
```

automates/program_analysis/GCC2GrFN/gcc_ast_to_cast.py
```diff
@@ -514,12 +520,49 @@ def parse_conditional_statement(self, stmt, statements):
...
+                loop_block = self.loop_exits[loop_index]
+                if target >= loop_block[0] and target <= loop_block[1]:
+                    if loop_block[0] > curr[0] and loop_block[0] < curr[1]:
```
dpdicken, line 557: Here, couldn't you just check 

```
if target == loop_block[1]:
    return [ModelBreak()]
```

and then remove this conditional and the one on line 560?

dpdicken (changes requested):

PR #266, opened 2021-10-20
Improvements to parsing metadata links 

dpdicken:
* Adds and improves scripts for adding metadata into GrFN
* Changes to parse new structure of alignment data
* Adds Unit metadata

TODO: Still needs a fix to handle changes from https://github.com/ml4ai/automates/pull/263 . Essentially, text spans are now a list of spans instead of a single span and should be handled appropriately.


titomeister (approved):
Looks good to me!!!

PR #271, opened 2021-11-02
Debugging test t3f

maxaalexeeva:
main change: drop Identifier label on an expanded Identifier mention---it is no longer just an identifier if we expand---this was what broke test t3f

other:
- added det to invalid outgoing to one of new invalid outgoings---it was expanding to just the determiner;
- set hops back to five; 
- remove 'foundBy' in toAttachment---key was not found because foundBy does not occur in all types of attachments, which broke align test


maxaalexeeva:
@alicekwak it should be safe to merge this into your branch (even if tests fail here). I can walk you through the changes tomorrow or Wednesday if something is unclear.

alicekwak (approved):
Thank you for debugging the test! I think I understood what you did. We can briefly discuss this during our weekly TR meeting, if needed. Thanks once again for solving this issue. It has been really bothering me 🥲

PR #272, opened 2021-11-02
fixing serialization of cross sent mentions

maxaalexeeva:


alicekwak (approved):
Wow, you've made a lot of changes in the test. I'll merge this and see how it works. Thank you so much for looking into this! :)

PR #275, opened 2021-11-15
[PA]: Allow custom gcc-10 path and more if-statement unit tests for gcc plugin

rsulli55:
This PR is consists of two things:

1) Updating `run_gcc_to_grfn.py`, the gcc plugin Makefile, and `test_gcc_plugin_c.py` with the goals of
  - making it possible to provide a custom gcc-10 path
  - simplifying the creation of the `json` files required for writing unit tests (only one invocation of `run_gcc_to_grfn.py` is needed)
 These goals were achieved by adding additional command line arguments to `run_gcc_to_grfn.py` and extending the Makefile and `test_gcc_plugin_c.py` to check the environment for a `CUSTOM_GCC_10_PATH` variable. 
 
 2) Updating the `if_statement` unit test, and writing the `if_else_statement` and `if_elif_statement` unit tests.  There are some issues with the creating CAST/GrFN for these unit tests which I outline below so I have marked them `skip` for now.
 
 ### `if_statement` unit test
 I extended this unit test as we discussed so that the variable `a` is referenced in both `if` statements.   After further discussion, I made it so that `a` is not only assigned, but accessed in the second `if` statement.  However, the GrFN is not what we expected (there is no arrow from `a` at the top to the `I` block in the second `if` statement).  The CAST looks fine though.  Furthermore, the created GrFN cannot be executed.
**See Issue #277**
[if_statement--CAST.pdf](https://github.com/ml4ai/automates/files/7550501/if_statement--CAST.pdf)
[if_statement--GrFN.pdf](https://github.com/ml4ai/automates/files/7550502/if_statement--GrFN.pdf)

### `if_else_statement` unit test
This unit test also has problems with the created GrFN.  The issues again appears to be the `I` blocks, and as above, the CAST looks fine.
**See Issue #276**
[if_else_statement--CAST.pdf](https://github.com/ml4ai/automates/files/7535081/if_else_statement--CAST.pdf)
[if_else_statement--GrFN.pdf](https://github.com/ml4ai/automates/files/7535083/if_else_statement--GrFN.pdf)

### `if_elif_statement` unit test
This unit test has problems with the CAST (which also causes the GrFN to be incorrect and execution results to be wrong).  In the CAST you can see the `else if` statements are not attached to the `else` block like they should. I checked the `gcc_ast.json`, and that correctly attaches the `else if` off the `else` branch.
**It sounds like this issue is fixed on a branch Tito is working on**
[if_elif_statement--CAST.pdf](https://github.com/ml4ai/automates/files/7535093/if_elif_statement--CAST.pdf)
[if_elif_statement--GrFN.pdf](https://github.com/ml4ai/automates/files/7535094/if_elif_statement--GrFN.pdf)


titomeister (approved):
Looks good!

PR #281, opened 2021-12-19
Unit and param men fix dec2021

maxaalexeeva:
- remove param setting mentions if there are more complete, overlapping interval param settings 
- disallow ParamAndUnit mentions that have more than one identifier (each mention should be associated with only one identifier)
- added a copyWithFoundBy method (for easier labeling of mentions with actions involved in their creation), e.g., here, the mention has been produced by the `var-word-equals_value-token` and the `assembleVarsWithParamsAndUnits` action:
```
List(ParamAndUnit, Event) => refugee move speed is equal to 200 
     ------------------------------ 
     Rule => var-word-equals_value-token++assembleVarsWithParamsAndUnits 
     Type => RelationMention 
     ------------------------------ 
     value (Value, Measurement, Phrase, Entity) => 200 
     unit (Unit, Measurement, Phrase, Entity) => km per day 
     variable (Phrase, Entity) => refugee move speed 
```
- removed ValueAndUnit (not being used)

### DIFF
--- a/automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/taxonomy.yml
+++ b/automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/taxonomy.yml
@@ -11,8 +11,7 @@
       - CommLineParameter
       - Measurement:
         - Value
-        - Unit:
-          - UnitRelation
+        - Unit
       - Filename
       - Language
       - Repository
@@ -37,3 +36,4 @@
   - Context
   - CommandSequence
   - MD_Context
+  - UnitRelation
--- a/automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/taxonomy.yml
+++ b/automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/taxonomy.yml
@@ -10,8 +10,7 @@
       - Parameter
       - CommLineParameter
       - Measurement:
-        - Value:
-          - ValueAndUnit
+        - Value
         - Unit:
           - UnitRelation
       - Filename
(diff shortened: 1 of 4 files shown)


automates/text_reading/src/main/resources/org/clulab/aske_automates/grammars/taxonomy.yml
```diff
@@ -10,8 +10,7 @@
       - Parameter
       - CommLineParameter
       - Measurement:
-        - Value:
-          - ValueAndUnit
+        - Value
         - Unit:
           - UnitRelation
```
alicekwak, line 15: Would you be able to move `UnitRelation` from Entity to Event?

alicekwak (changes requested):
This is great, thank you @maxaalexeeva! Would you be able to move "UnitRelation" from Entity to Event? Except for that, everything seems perfect :)

alicekwak:
Oops, and it looks like one ParameterSetting test (u5a) is failing. The test sentence is this: 

> For the purposes of model parameterisation the value of shoot_lag has been assumed to be around 40 o Cd.

It's failing because "value of shoot_lag" is captured as a variable instead of "shoot_lag". Do you have any idea why this change occurred?

maxaalexeeva:
@alicekwak 

> It's failing because "value of shoot_lag" is captured as a variable instead of "shoot_lag". 

Nooooo! That was one of those accidentally fixed tests and now it broke again because of who knows what! Bummer.

alicekwak (approved):
Thank you for the revision! Looks awesome :)

cl4yton (approved):
From what I understand, this looks good. Thanks!

PR #283, opened 2022-01-24
[PA]: CAST conditional updates

rsulli55:
- Processing of if statements and loops should be handled correctly in many cases that were previously broken
- The basic algorithm is to keep track of what CAST body a parsed basic block is attached to, and then when processing a new basic block attach it to be sibling of its parents nearest common dominator (dumped by gcc plugin)
- `do while` loops, `break`s, `continue`s, and labeled `goto`s will not be handled correctly by the algorithm.  a `do while` loop is currently indistinguishable (in the json) from a `while` loop with a `break` in it, which complicates things.
- Also updated one test to follow the format of other tests in `test_gcc_plugin_c.py`


jobagy (approved):
Reviewed all changes.

PR #285, opened 2022-02-04
Null code_span metadata GrFN generation

titomeister:
Added a small check in metadata.py that checks whether the input GrFN has no code_span metadata information. This is to allow GrFNs to be executed on the web API and to have their expression trees extracted in the case they have no code_span information.


cl4yton (approved):
Looks good!

PR #288, opened 2022-02-23
Updated and extended GCC test suites for new CAST spec

rsulli55:
### Changes to `gcc_ast_to_cast.py` and `cast.py` to accomodate testing
- To facilitate easier regeneration of test data updated `gcc_ast_to_cast.py` so that CAST node `SourceRef`s only use the file name and not the full path to the file.
- In `cast.py` changed the equality check on CAST objects so that `Name` nodes do not use the auto generated Swagger `__eq__` method.  The auto generated method will check the numerical id's of Name nodes, and we do not expect those to be constant across runs of CAST generation.  Instead, the equality check will just make sure the string names are the same.
###  Updating old tests for GCC plugin to GrFN
- Renamed `test_gcc_plugin_c.py` to `test_gcc_plugin_to_grfn_c.py` because it tests all the way to GrFN execution
- Moved test data for these tests to a central directory.  This makes it easier to regenerate test data if needed.
- Regenerated all test data to accommodate the new CAST spec.
### Added new tests
- Based on the test cases we have collected over the last few months I added two new test suites, one for testing the GCC plugin (C file to AST json) and one for testing `gcc_ast_to_cast.py` (AST json to CAST).
- Extending these test suites with more test cases will be very easy thanks to `pytest`s parameterize functionality.  To add more tests, all that is needed is to create the two test data files (C and AST json or AST json and CAST json) and then place these files in the corresponding test data directory.  Tests for these file will automatically be added to the test suite.


jobagy (approved):
Reviewed all changes and they look good.

PR #291, opened 2022-04-28
Annotated CAST

rsulli55:
Merging Annotated CAST to master.  The legacy CAST -> AIR -> GrFN pipeline is not removed.  We added a flag to `gcc_ast_to_cast.py` so that "legacy" CAST generation is possible.  Legacy CAST has a simple change: the `FunctionDef'`s `name` attribute is a string instead of a `Name` node.  This legacy CAST should be used for the legacy GrFN pipeline.  The AnnCAST pipeline requires `FunctionDef` nodes to have `Name` nodes for the `name` attribute.


jobagy (approved):
Ryan and I worked on this branch together over the last few months. This merge allows us to continue with the work for the remaining list of issues but also allows running the legacy GrFN pipeline. I reviewed all files except for the generated text case json files.

PR #294, opened 2022-05-05
AnnCast-To-GrFN: Updates for 2.3 FunctionDefs/Calls and General Fixes

rsulli55:
- Implemented a flag `FROM_SOURCE_FOR_GE` which will change the from source metadata for VariableNodes to be as they requested
- Specialized globals are now used for 2.3 FunctionDefs
- Resolved a GrFN generation issue with Interface nodes that are neither grandparents nor have a grandparent
- Added variable ids to variables in lambda expressions to fix an issue with variable name collisions
- Changed loop container's top interface to be of type `LoopTopInterface`.  This will be helpful later during execution implementation
- Added in dummy assignments to FunctionDef's when those variables are expected to propagate through a container's top interface.  This resolves issues for CAST generated from Python due to dynamic variable creation e.g.
```python
def func():
    z = 3
    if z == 3:
        y = 5
    else:
        y = 7
```
The AnnCast pipeline expects `y` to enter through the If container's top interface, so we add a dummy assignment `y = None` at the top of `func()`

### DIFF
--- a/automates/model_assembly/metadata.py
+++ b/automates/model_assembly/metadata.py
@@ -425,6 +425,13 @@ class VariableCreationReason(AutoMATESBaseEnum):
     CONDITION_RESULT = auto()
     LOOP_EXIT_VAR = auto()
     LITERAL_FUNCTION_ARG = auto()
+    TOP_IFACE_INTRO = auto()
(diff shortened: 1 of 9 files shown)


jobagy (approved):
Changes are excellent.

PR #295, opened 2022-05-11
[PA]: Testing Updates

rsulli55:
- added AnnCast testing suite
- added GrFN testing suite for AnnCast pipeline
- refactored and added more tests to gcc to cast pipeline


jobagy (approved):
Changes look good.

PR #300, opened 2022-05-13
PA: Code Cleanup and Reorganization

rsulli55:
- Moved annotated CAST helper functions to a separate file
- Resolved TODOs in the AnnCast code
- Removed extraneous files
- Renamed things for clarity such as AnnCast to PipelineState
- Made a separate testing suite for GrFN 2.2
- Updated scripts and added additional command line arguments
- Cleaned up imports


jobagy (approved):
All looks good. One note, since we moved debugging prints under flag, git is marking those as not covered. This will decrease code coverage stats.

PR #301, opened 2022-05-21
Claytonm/requirements

cl4yton:
Hi @titomeister : This has been a task I have meant to get to for a long time. Dependabot has identified several of our requirements.txt module versions that are at older versions that should be increased. This PR makes the following changes to requirements.txt:

- notebook from 6.4.1 to 6.5.10
- numpy from 1.18.2 to 1.21
- nltk from 3.4.5 to 3.6.6

nltk is only used by scripts/code_summarization , which is older code created by Paul Hein. I've added to the README here that this code may need to be verified, but I'm ok bumping without testing for now.

I'm pretty sure both of the notebook and numpy changes will not cause problems, but could you please run our PA pipeline and its tests to verify?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,9 +6,9 @@ flask_wtf==0.14.3
 future==0.18.2
 matplotlib==3.3.3
 networkx==2.4
-nltk==3.4.5
-notebook==6.4.1
-numpy==1.18.2
+nltk==3.6.6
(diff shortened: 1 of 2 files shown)


titomeister (approved):
The tests look like they're passing on the CI pull request. I've also checked them by hand with the branch and it looks normal to me as well.

PR #303, opened 2022-05-31
Increasing versions of setup.py modules that currently have security vulnerabilities

cl4yton:
Hi @titomeister ,

This is related to the previous PR that bumped up the versions of several modules in requirement.txt -- I forgot that of course there will be similar dependencies in setup.py. This PR addresses that.

Increasing versions for notebook to 6.4.10, numpy to 1.21 and nltk to 3.6.6.

Can you please check that these increases in setup.py don't appear to cause any issues, and if so, then please go ahead and squash & merge into master.


titomeister (approved):
The tests appear to be passing with no issues given these changes. Github isn't reporting any issues with the CI. Everything looks good to me!

PR #315, opened 2022-09-02
GroMEt USub Pull Request

titomeister:
Pull Request to merge in changes that add support for Unary Subtraction (USub) GroMEt generation.

Also fixes a minor bug with creating pofs when it comes to Binary Operations.


cl4yton (approved):
Thanks @titomeister , this looks good; and it was a much smaller adaptation than I expected (which is a good thing).

PR #316, opened 2022-09-03
GroMEt Dictionaries PR

titomeister:
Adds initial support for GroMEt dictionary generation:

In particular:
 - Generating CAST Dictionary LiteralValue 
 - Visualization for Dictionaries
 - Generating a GroMEt dictionary literal
 - Generating _get and _set calls for dictionaries.

One change that was done was to the '_List_get' and '_List_set' functions. They're currently changed to be type-general, in other words they're generic '_get' and '_set'. In the future this will probably change to be more type-specific.


cl4yton (approved):
Looks great @titomeister , thanks!

PR #318, opened 2022-09-09
CAST Version 1.2.1 PR

titomeister:
This pull request updates the GroMEt structures

GroMEt metadata
- The metadata collection has been changed such that all metadata is stored in an array of lists, where each list holds metadata for a particular GroMEt object. The GroMEt objects then store indices into this list now, as opposed to storing the metadata itself.
- The exporter and importers were updated to support this change
- One temporary fix needed to be introduced: Swagger's to_dict() methods don't seem to fully support array of lists, so an additional step is introduced in the serializer to make sure all objects get converted to dictionaries before serializing. 
- A schema and schema_version field were introduced to the top level GroMEt structure.

There were plans to introduce an identifier in CAST to maintain an execution order, but this has been pushed back until later.


vincentraymond-ua (approved):
Thanks @titomeister for making this request, it looks good to me! Nothing is immediatly jumping out.

PR #320, opened 2022-09-15
GroMEt For Loop PR

titomeister:
Implements changes to GroMEt generation to support for loop generation for Python.

- Makes use of the 'init' field in GroMEt FNs that have a loop. 
- Introduces functions to interact with iterators ('next', 'init') in CAST.
- Fixes some other minor bugs that prevented GroMEts for simple expressions and functions to be generated.
- Fixed the metadata indexing issue.
- Improves how the primitive generation is handled, though more work is needed to make it fully complete. We now keep track of how many inputs/outputs each primitive uses, which is needed for generating GroMEts of primitive functions.


vincentraymond-ua (approved):
Looks good @titomeister, this has definitely helped my understanding of how some of the Gromet wiring is done

PR #323, opened 2022-10-01
GroMEt record initial implementation PR

titomeister:
(NOTE: The majority of the files changed in this PR are CAST files that are auto-generated by swagger)
This PR implements the addition of an initial Record implementation in GroMEt.
The Record is a general definition for Python Classes, C structs, and other language's representation of Objects.

For now the Record implementation has only been tested on Python.

- Modifies to_gromet_pass.py to add GroMEt generation for RecordDef types. Requires more optimization
- Removes ClassDef CAST/AnnCAST nodes and replaces it with RecordDef CAST/AnnCAST nodes. 
- Begins adding support for Attribute nodes as these are heavily used in Python's objects system.
- Adds support to the other six AnnCAST passes to support RecordDef and Attribute.
- Adds ModelImport CAST/AnnCAST node, even though it isn't used yet by this branch.

NOTE: The id system for variables and functions related to classes is currently wrong, but it isn't used by the GroMEt generation, its main use is for GrFN, so it won't affect this initial RecordDef generation. It will need to be fixed soon, though.


cl4yton (dismissed):
This looks great!

cl4yton (approved):
Hi @titomeister , I finished the review and these updates look good.

PR #324, opened 2022-10-07
Gromet Import PR

titomeister:
This branch contains a great amount of changes/additions to the GroMEt pipeline.
- Support for attributes was rewritten in CAST, and expanded a little in GroMEt. 
- Some modifications for AnnCAST visitors in order to support attributes and importing.
- Import support was updated to better reflect the current way we're handling Imports in GroMEt. We generate CAST for each individual import, and use these to generate GroMEt ImportReferences.
- Adds some helper support for determine when functions are from imported modules or builtins. 
- There's still a great deal of the Bucky model that needs to be properly ingested, so code was added to 'guard' against unimplemented features breaking the pipeline. 

One known issue is there's a discrepancy with the loops when generating GroMEt for other models. We're aware of this and working on a fix.


vincentraymond-ua (approved):
Looks good @titomeister! 

I've confirmed that the whole pipeline (py -> CAST -> Gromet -> JSON -> Gromet) is working for the simplified Bucky model

PR #329, opened 2022-11-18
Claytonm/gromet query

cl4yton:
Adds a cheap-n-cheerful simple query to grab all named output ports, which correspond to named variables that get set to values (note, this does not include instances of named *input* ports, which would be for named arguments to functions). 

The function `collect_named_output_ports` does this, and is defined in a new file `<automates_root>/automates/gromet/query/query.py`.

The expectation is that this functionality should be subsumed by the rust gromet graph query interface.


vincentraymond-ua (approved):
@cl4yton - Looks good on my end for Gromet 0.1.4! I still need to add the new Metadata types to the JSON2Gromet importer and then Gromet 0.1.5 will be fully supported, and I can submit a pull request to have that merged.

PR #333, opened 2022-12-12
Gromet FN/Metadata v0.1.5 support for Gromet Importer

vincentraymond-ua:
Updates the json_to_gromet importer to support the v0.1.5 format for both Gromet FN and Gromet Metadata. 

Additionally, updates the process_file_system function to be more functional by returning the Gromet module collection and disabling writing to file unless write_to_file argument is True.


automates/program_analysis/JSON2GroMEt/json2gromet.py
```diff
@@ -16,64 +17,83 @@
...
-
 def json_to_gromet(path):
     # Read JSON from file
     with open(path) as f:
         json_string = f.read()
     json_object = json.loads(json_string)
 
-    # Creat Module object
-    gromet_module = GrometFNModule()
+    # Create top level module collectio
```
kwalcock, line 47: collection?

kwalcock:
I'm concluding that the specs from files like https://github.com/ml4ai/automates-v2/blob/master/docs/source/gromet_FN_v0.1.5.yaml are really only for human consumption.  They are there so that others can manually write code to process the format.  That seems unfortunate.  If correct, I would remove the discussion of code generation.

cl4yton:
Hi @kwalcock : Thanks for your comment -- I want to make sure I understand. 

Some background (which you may already know, but in case I didn't review with you...): It is the case that we currently use the swagger-codegen generated data model (actually three: CAST, GrometFN, and Metadata) in the Program Analysis (PA) pipeline. The auto-generated code ends up being directly versioned within the automates repo in these locations (all under `<automates_root>/automates/`):

- CAST: `.../program_analysis/CAST2GrFN/model/`  (at some point this will be moved to a better location)
- GrometFN: `.../gromet/fn/`
- Metadata: `.../gromet/metadata/`

Inserting the swagger-codegen generated code into these respective locations does require some post-processing to fix paths. I created a script that manages this: <automates_root>/scripts/swagger/codegen_swagger_models.py . For each of the three models (CAST, GrometFN, Metadata), this script 

1. Copies the existing corresponding data model directory in automates into a temp directory; this makes it easy to "roll back" if something goes wrong with the running of the script; if all goes well, this temp directory can be manually removed.
2. Runs swagger-codegen for the model (this uses the swagger-codegen command in step 2 of the instructions in the preamble of the yaml model files).
3. Copies the generated model files into the corresponding data model directory, adjusting paths as needed in each file to match the directory context.

Now, I believe you are referring to the "preamble" text that is at the start of each of the swagger model yaml files that describes how to use swagger-codegen to generate the source code class structure. These instructions were originally created as a reminder to ourselves for how these are generated (the original version of these instructions were created by Daniel Dicken when he first created the swagger spec). Until I had created the above script, I would perform the generation and insertion manually. Since creating the script, I have maintained the instructions at the top for each model spec, in case others who might wish to programmatically interact with the data models but not using automates want to do so. But within automates, we use the script described above to maintain the data model.

@vincentraymond-ua maintains a serializer/deserializer for the internal automates data structures based on the GrometFN (in particular, the GrometFNModel and GrometFNModuleCollection) and Metadata, where the export target is a particular JSON format that we call the "Gromet FN relational database schema". This JSON format is governed by ASKEM "gromet schema" conventions that don't exactly match the swagger model, which is why we maintain the serializer/deserializer and don't just directly use the build-in `.to_json` and `load`. It is a bummer that we can't just directly do things that way.

So, this PR is about Vincent updating the serializer/deserializer to support the latest changes to the Metadata data model (v0.1.5).

Now that we have the explicit internal data structure (using the GrometFN and Metadata data structures derived from the swagger models), along with our support for the serializer/deserializer, I would recommend using those to programmatically interact with the Gromet FN model, as it is generated by the Program Analysis function network extraction pipeline. Justin Lieffers is also currently developing a graph database representation which is suited for querying and manipulating function networks, and this is created directly from the Gromet FN model in the automates PA pipeline. But I think the instructions for how the data models are generated directly using swagger-codegen might also still be of use to some.

Does this help clarify what you are responding to in your comment? I do think that the way we have things documented is leading to confusion. Given the above context, what would your recommend for making this clear? (... And are there other issues we should consider?)

Thanks!

PR #336, opened 2022-12-16
Updating minimum Python version to 3.9.0

vincentraymond-ua:
Updating the minimum required Python version in setup.py and environment.py to 3.9.0.


titomeister (approved):
The version changes look good!

